PR #43, opened 2014-04-05
Add root device size option

victorlin:
For some of our kitchen usage, it took more than 8 GB to run. We need to adjust the size of root device size so that it can run. This pull request provides an extra option `root_device_size` which allows you adjust the root device size.


tyler-ball:

PR #44, opened 2014-04-16
add retry on request-limit-exceeded error

alexism:
When running TestKitchen with many suites in parallel, the requests to EC2 might get throttled, causing TestKitchen to fail.

This pull request adds a retry mechanism on `create_server` and `destroy_server`.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -16,6 +16,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'kitchen/driver/retry'
+
 require 'benchmark'
 require 'json'
 require 'fog'
@@ -29,6 +31,8 @@ module Kitchen
     #
     # @author Fletcher Nichol <fnichol@nichol.ca>
     class Ec2 < Kitchen::Driver::SSHBase
+
+      include Kitchen::Retry
 
       default_config :region,             'us-east-1'
       default_config :availability_zone,  'us-east-1b'
@@ -84,8 +88,10 @@ module Kitchen
       def destroy(state)
         return if state[:server_id].nil?
 
-        server = connection.servers.get(state[:server_id])
-        server.destroy unless server.nil?
+        with_retry_on_throttling do
+          server = connection.servers.get(state[:server_id])
+          server.destroy unless server.nil?
+        end
         info("EC2 instance <#{state[:server_id]}> destroyed.")
         state.delete(:server_id)
         state.delete(:hostname)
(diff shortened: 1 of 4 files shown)


tyler-ball:
Some notes from @brycebignasty
----
When running kitchen converge or test with --concurrency greater than 10, one or more of the VMs will usually not be created.  When I try the same concurrency with kitchen create, I get this exception:
```
>>>>>> ------Exception-------
>>>>>> Message: Failed to complete #create action: [Request limit exceeded.]
>>>>>> ----------------------
```
I have a workaround which I think also proves the problem:  In my Thorfile I have separated the destroy/create/converge/verify/destroy steps into separate commands, and run the create and destroy steps with --concurrency=10, with the rest at full concurrency.

I'm running 15 suites concurrently, with one of the instances having an extra EBS volume.

I don't know if this issue belongs here, or in the test-kitchen project itself.

Thanks,
-Bryce.

tyler-ball:
Anytime we use high concurrency to create/destroy servers we are going to run into this error.  One potential solution is to use batch requests of servers, but with the current threading model in the concurrent case I don't think that is realistic.

Test Kitchen uses a thread pool when running its concurrency.  Something we could do is lower the threading pool size default when running against AWS to avoid hitting the API limits.  Users could still overwrite this at their own danger (or if they write a custom driver which does batch requests).

If you want to change your threading size now you can set `concurrency: 10` in your .kitchen.yml (as a root level config, not under `driver` or anything).  Otherwise it defaults to the number of threads to the number of cases.

bignastybryce:
I think an optimum solution would allow for throttling create/destroy calls (and the associated status calls) while still allowing converge/verify to run as many threads as there are suites.  -c 10 on my current build takes it from 21 minutes (using the workaround I mentioned in my original issue) to 32 minutes, which is a bummer.  I'm trying to go the other way.

i11:
+1 PR
"...in short, this is adding 62 new lines of code for a problem that affects only a small percentage of people..." is inaccurate statement in my opinion.

fhanson:
+1 this would be an extremely helpful addition to kitchen-ec2

cheeseplus:
I think the idea behind the PR might still be worth exploring but given the time since it was written the this driver has changed a great deal. We've moved off Fog and have included the gem `retryable` which could handle some of the duties here as we already use it for tagging https://github.com/test-kitchen/kitchen-ec2/blob/master/lib/kitchen/driver/ec2.rb#L201. I'm going to close this one for these reasons but if someone wanted to resurrect against current code I'd happily give it a look.

PR #53, opened 2014-05-16
refresh ec2 hostname/ip from server id 

angrycamel:
This change will refresh the EC2 instance hostname/ip immediately prior to the first ssh connection setup for the driver lifetime.

I see test-kitchen used in 2 complementary ways - full CI on the one hand, and developer converge/verify hacking on the other. This change is targeted at the latter. 

The change solves a problem where (to save $$$) the developer stop the EC2 instance before going home and start it again in the morning to resume work. Because EC2 assigns a new public ip when the instance is restarted, it would break kitchen-ec2 because the public ip stored in the state file would never be updated. 

This change doesn't fix the state file not being updated - so 'kitchen diagnose' will show incorrect information. Fixing this would need a change to test-kitchen/test-kitchen.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -65,6 +65,11 @@ module Kitchen
       required_config :aws_secret_access_key
       required_config :aws_ssh_key_id
       required_config :image_id
+ 
+      def initialize(config = {})
+        super
+        @hostname_refreshed = false 
+      end
 
       def create(state)
         return if state[:server_id]
@@ -99,6 +104,11 @@ module Kitchen
 
       def default_username
         amis['usernames'][instance.platform.name] || 'root'
+      end
+
+      def build_ssh_args(state)
+        refresh_host_if_needed(state)
+        super
       end
 
       private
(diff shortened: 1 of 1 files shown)


tarak:
Would this also help when assigning an elastic IP to an EC2 instance? When I do so in my cookbook I can successfully converge the node but not login later on. I need to change the IP or host name in the instance's YML file at

```
.kitchen/default-ubuntu-1404.yml
```

jkeiser:
I love this!  I think we need to make sure we don't re-get the information the first time we run kitchen though--this is only for the second time.

zl4bv:
Will this work for non-SSH based transports such as WinRM? It looks like the version of this class still inherits from the deprecated SSHBase class.

cheeseplus:
If we can get this rebased we're happy to look at it, otherwise we'll close in a month from now.

cheeseplus:
It's a bit longer than a month but going to close given that it's inheriting from the deprecated SSHBase Class. Definitely a great addition that we'd love to have if you can do a new pull request in the future.

PR #64, opened 2014-08-19
Add root device size updated

rvalyi:
an updated version of https://github.com/test-kitchen/kitchen-ec2/pull/43 by @victorlin


tyler-ball:

Issue #55, opened 2014-05-21
Use IAM role to authenticate with AWS


rayrod2030:
It would be nice to be able to have kitchen-ec2 create ec2 instances using permissions derived from an IAM role.  This feature would be useful in situations where test-kitchen/kitchen-ec2 is being run on a jenkins CI server in EC2 which is a common setup since you cannot run vagrant on top of a hypervisor.

The underlying aws library fog already has this capability in the form of the use_iam_profile options attribute so it would just be a matter of passing along this flag as well as removing the need to make the aws access key and secret a required config.

Just wanted to see if anyone else thinks having the ability to use IAM roles to create EC2 test-kitchen instances would be useful.

PR #68, opened 2014-09-10
Add support for IAM credentials

daanemanz:
As mentioned in https://github.com/test-kitchen/kitchen-ec2/issues/55.

This change will add support for using IAM temporary credentials when creating EC2 instances. Currently the credentials will have to be set in the environment, while an EC2 instance that's set up with an IAM profile can fetch its temporary credentials from the metadata server.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -32,7 +32,7 @@ module Kitchen
 
       include Fog::AWS::CredentialFetcher::ServiceMethods
 
-      iam_credentials = fetch_credentials(use_iam_profile: true) rescue {}
+      iam_creds = fetch_credentials(use_iam_profile: true) rescue {}
 
       default_config :region,             'us-east-1'
       default_config :availability_zone,  'us-east-1b'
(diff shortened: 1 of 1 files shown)


Igorshp:
:+1:  Was just looking into enabling this. Glad someone already fixed the problem.
inability to use roles for auth is the only thing stoping kitchen to be run on jenkins at the moment

willejs:
:+1: this looks good. :shipit: :ship:

tyler-ball:
`iam_creds` shouldn't be a class-level local variable.  It can be a memoized instance variable like
```ruby
def iam_creds
  @iam_creds ||= begin
    fetch_credentials(use_iam_profile: true)
  rescue # TODO what exception does this raise on failure?
    debug("fetch_credentials failed with exception #{e.message}:#{e.backtrace.join("\n")}")
    {}
  end
end
```

I also don't like rescuing all possible exceptions.  That could be catching a `SystemExit` or something and hiding good debugging information from us.  Otherwise, this looks good to me.

tyler-ball:
Closing this as it was merged in #104

PR #76, opened 2014-10-30
Allow EBS volume type to be specified.

duggan:
Allows SSD backed volumes to be used. Speed++


tyler-ball:

PR #78, opened 2014-11-06
Added config option and method to allow specifying block device mappings

nsdavidson:
Added the ability to specify block device mappings on EC2 instances.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -72,6 +72,8 @@ module Kitchen
       required_config :aws_ssh_key_id
       required_config :image_id
 
+      default_config :block_device_mappings, nil
+
       def create(state)
         return if state[:server_id]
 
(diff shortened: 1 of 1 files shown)


tyler-ball:
Thanks for taking a stab at this - I like the idea, but I have a couple todos before we can merge this:

1. We need to add documentation to the README explaining this new config value
1. Now that we are adding the `block_device_mappings` config, I think we should get rid of the 3 old values (`ebs_volume_size`, `ebs_delete_on_termination` and `ebs_device_name`) and have users populate the `block_device_mappings` config with the hash of values they want.
1. Because we want to be backwards compatible, we need to mark the old configs as depreciated in the README and point people to the new configs.  It should also introspect the test-kitchen version and raise an error if the version is >= 2.0.0 reminding us to remove these deprecated configs.
  1. To support backwards compatibility we will need to accept the old keys if they are provided.  To encourage switching to the new value we could also print out a warning message.  If they provide both the old keys and the new keys, we can raise an exception and point them to using the new config key.

Thanks for your work on this!  Let me know if you have any issues with this or you need help.

tyler-ball:

Issue #86, opened 2014-12-30
The plugin does not create /etc/chef/ohai/hints/ec2.json file


dtoubelis:
It was stated in ohai discussion that reliably detecting cloud info is very difficult and it seems that common practice now create a file in /etc/chef/ohai/hints directory.

Could you please create an empty `/etc/chef/ohai/hints/ec2.json` file after chef client is installed?

Here are relevant discussions:
- https://tickets.opscode.com/browse/OHAI-267
- https://tickets.opscode.com/browse/CHEF-3160

tyler-ball:
I'm going to close this since it is included in https://github.com/test-kitchen/kitchen-ec2/pull/89 and will be merged with that fix.

PR #87, opened 2014-12-30
Create /etc/chef/ohai/hints/ec2.json

dtoubelis:
This should create `/etc/chef/ohai/hints/ec2.json` during create stage as stated in #86. I'm new to Ruby and I couldn't test it properly, so please double check before merging. It would also be good and relatively easy to create a test case for ths but I'm complete noob, I have no idea so far how to do it.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -103,6 +103,10 @@ module Kitchen
           :ssh_timeout => config[:ssh_timeout],
           :ssh_retries => config[:ssh_retries]
         })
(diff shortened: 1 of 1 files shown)


rmoreira:
Great fix.
I've been having the same issue with Chef 12

dtoubelis:
Writting to '/etc' require root priveleges, so 'sudo' should do the trick.

tyler-ball:
I'm going to close this since it is included in https://github.com/test-kitchen/kitchen-ec2/pull/89 and will be merged with that fix.

PR #90, opened 2015-01-29
Fix security_group_ids parameter for spot requests

gfloyd:
The parameter used by Fog to create a spot request is actually `:groups`. This fixes the parameter so that specifying security groups when requesting spot instances with the `price` option works.


tyler-ball:
Verified this locally - thanks for the fix!

Issue #71, opened 2014-10-03
Can't find how to set EBS Volume Size


sogos:
Hello, I can't find how to set EBS Volume size. It's always 1Gb

Here my conf

```ruby
---
provisioner:
  name: chef_solo

driver:
  name: ec2
  aws_access_key_id: XXXXX
  aws_secret_access_key: YYYYYYY
  region: eu-west-1
  aws_ssh_key_id: GER
  availability_zone: eu-west-1a
  security_group_ids: ["sg-XXXXXX"]
  ssh_key: /root/XXXX.pem
  require_chef_omnibus: true

platforms:
  - name: debian6
    driver:
      image_id: ami-XXXXXX
      username: root
      ebs_volume_size: 40,
      ebs_delete_on_termination: 'true'
      ebs_device_name: '/dev/sda1'

suites:
  - name: default
    run_list:
      - recipe[ger::default]
    attributes:

tyler-ball:
@sogos Is this still a problem for you?

PR #98, opened 2015-02-16
New `block_device_mappings` config

tyler-ball:
This PR rolls a bunch of existing, storage based PRs into one:

* Closes https://github.com/test-kitchen/kitchen-ec2/issues/71
* Connects to https://github.com/test-kitchen/kitchen-ec2/pull/64
* Connects to https://github.com/test-kitchen/kitchen-ec2/pull/43
* Connects to https://github.com/test-kitchen/kitchen-ec2/pull/78
* Connects to https://github.com/test-kitchen/kitchen-ec2/pull/76

I am deprecating the old, individual storage config keys and adding a new one which accepts a list.  This allows users to specify multiple block device mappings.  If one of the entries has the same name as the image root storage device it will be overridden.

To support backwards compatiability if you do not use the new `block_device_mappings` config it will be auto created using the old deprecated configs.  It emits a warning because we will eventually remove this functionality.

How are the READMEs?  What do you think about usability?

\cc @fnichol @rvalyi @victorlin @nsdavidson @duggan

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,15 +112,45 @@ or `nil` otherwise.
 
 ### <a name="config-ebs_volume_size"></a> ebs\_volume\_size
 
-**Required** Size of ebs volume in GB.
+**Deprecated** See [block_device_mappings](#config-block_device_mappings) below.
+
+Size of ebs volume in GB.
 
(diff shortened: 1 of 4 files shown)


tyler-ball:
Okay tests have been added - this is ready for review!

PR #89, opened 2015-01-27
Fixing issue with logging in with private interface and the server never being ready to log in

rmoreira:
This fixes logging in to the server without using public ip address

Fixes https://github.com/test-kitchen/kitchen-ec2/issues/86

PR #102, opened 2015-02-23
Adding ohai hints

tyler-ball:
From https://github.com/test-kitchen/kitchen-ec2/pull/89 - I meant to add these a while ago and simply forgot

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -128,6 +128,7 @@ module Kitchen
           :ssh_retries => config[:ssh_retries]
         })
         print "(ssh ready)\n"
+        create_ec2_json(state)
         debug("ec2:create '#{state[:hostname]}'")
       rescue Fog::Errors::Error, Excon::Errors::Error => ex
         raise ActionFailed, ex.message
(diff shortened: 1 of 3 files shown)


lib/kitchen/driver/ec2.rb
```diff
@@ -341,6 +342,15 @@ def block_device_mappings
 
         bdms
       end
+
+      # TODO Update this to use the new Transport logic when that is ready
+      def create_ec2_json(state)
+        Kitchen::SSH.new(*build_ssh_args(state)) do |conn|
+          run_remote("sudo mkdir -p /etc/chef/ohai/hints;sudo touch /etc/chef/ohai/hints/ec2.json",
```
jtimberman: Will sudo always be available? I think we use it elsewhere but I may misremember.

tyler-ball: This is not an ideal fix - when we can leverage the Transport logic from the windows-guest-support branches we will move this logic there.  This is just a temporary workaround until we get access to that.

Having said that, do you think I should add a feature flag or something, so people on systems without sudo (or `touch`) aren't totally borked by this?

fnichol:
@tyler-ball Will this need to be reworked to fit in with #109? The only thing making me wonder is the use of `Kitchen::SSH` which can be swapped for the Transport code

tyler-ball:
Moving the changes into https://github.com/test-kitchen/kitchen-ec2/pull/109 and closing this

PR #104, opened 2015-03-10
Add support for IAM role credentials

Igorshp:
This is a patch for PR by @daanemanz - https://github.com/test-kitchen/kitchen-ec2/pull/68
The problem was that fetch_credentials does not exist in the context and would always fall back to rescue block.

The original issue is https://github.com/test-kitchen/kitchen-ec2/issues/55

PR #107, opened 2015-04-07
Add explicit option for using iam profile for authentication

JamesAwesome:
PR #104 broke the ability to use kitchen-ec2 with explicit aws credentials set via .kitchen.yml while on an ec2 host.

When on an ec2 host the `#fetch_credentials` method which is included from `Fog::AWS::CredentialFetcher::ServiceMethods` will always return a hash with a value set for the `:aws_session_token` key. Because of this `config[:aws_session_token]` will always default to `iam_creds[:aws_session_token]`. Since `config[:aws_session_token]` is passed to `#connection`, `Fog::Compute` fails to authenticate using any credentials set in .kitchen.yml since the value of `:aws_session_token` is incorrect.

By adding an option to explicitly use a server's iam profile we can allow hosts which live in ec2 to use .kitchen.yml as well as their iam profile for driver configuration.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -29,7 +29,6 @@ module Kitchen
     #
     # @author Fletcher Nichol <fnichol@nichol.ca>
     class Ec2 < Kitchen::Driver::SSHBase
-
       extend Fog::AWS::CredentialFetcher::ServiceMethods
       default_config :region,             'us-east-1'
       default_config :availability_zone,  'us-east-1b'
@@ -41,14 +40,16 @@ module Kitchen
       default_config :private_ip_address, nil
       default_config :iam_profile_name,   nil
       default_config :price,   nil
+      default_config :use_iam_profile, false
       default_config :aws_access_key_id do |driver|
-        ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] || iam_creds[:aws_access_key_id]
+        ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] || driver.iam_creds[:aws_access_key_id]
       end
       default_config :aws_secret_access_key do |driver|
-        ENV['AWS_SECRET_KEY'] || ENV['AWS_SECRET_ACCESS_KEY'] || iam_creds[:aws_secret_access_key]
+        ENV['AWS_SECRET_KEY'] || ENV['AWS_SECRET_ACCESS_KEY'] \
+          || driver.iam_creds[:aws_secret_access_key]
       end
       default_config :aws_session_token do |driver|
-        ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN'] || iam_creds[:aws_session_token]
+        ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN'] || driver.iam_creds[:aws_session_token]
       end
       default_config :aws_ssh_key_id do |driver|
         ENV['AWS_SSH_KEY_ID']
@@ -98,10 +99,10 @@ module Kitchen
         end
       end
 
-      def self.iam_creds
+      def iam_creds
         @iam_creds ||= begin
-          fetch_credentials(use_iam_profile:true)
-        rescue RuntimeError => e
+          config[:use_iam_profile] ? fetch_credentials(use_iam_profile: true) : {}
+        rescue RuntimeError, NoMethodError => e
           debug("fetch_credentials failed with exception #{e.message}:#{e.backtrace.join("\n")}")
           {}
         end
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -49,8 +49,7 @@ module Kitchen
           driver.iam_creds[:aws_secret_access_key]
       end
       default_config :aws_session_token do |driver|
-        ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN'] ||
-          driver.iam_creds[:aws_session_token]
+        driver.default_aws_session_token
       end
       default_config :aws_ssh_key_id do |driver|
         ENV['AWS_SSH_KEY_ID']
@@ -175,6 +174,16 @@ module Kitchen
 
       def default_public_ip_association
         !!config[:subnet_id]
+      end
+
+      def default_aws_session_token
+        env = ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN']
+        if config[:aws_secret_access_key] == iam_creds[:aws_secret_access_key] \
+          && config[:aws_access_key_id] == iam_creds[:aws_access_key_id]
+          env || iam_creds[:aws_session_token]
+        else
+          env
+        end
       end
 
       private
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -40,7 +40,6 @@ module Kitchen
       default_config :private_ip_address, nil
       default_config :iam_profile_name,   nil
       default_config :price,   nil
-      default_config :use_iam_profile, false
       default_config :aws_access_key_id do |driver|
         ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] || driver.iam_creds[:aws_access_key_id]
       end
@@ -49,7 +48,7 @@ module Kitchen
           || driver.iam_creds[:aws_secret_access_key]
       end
       default_config :aws_session_token do |driver|
-        ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN'] || driver.iam_creds[:aws_session_token]
+        driver.default_aws_session_token
       end
       default_config :aws_ssh_key_id do |driver|
         ENV['AWS_SSH_KEY_ID']
@@ -101,7 +100,7 @@ module Kitchen
 
       def iam_creds
         @iam_creds ||= begin
-          config[:use_iam_profile] ? fetch_credentials(use_iam_profile: true) : {}
+          fetch_credentials(use_iam_profile: true)
         rescue RuntimeError, NoMethodError => e
           debug("fetch_credentials failed with exception #{e.message}:#{e.backtrace.join("\n")}")
           {}
@@ -166,6 +165,15 @@ module Kitchen
 
       def default_public_ip_association
         !!config[:subnet_id]
+      end
+
+      def default_aws_session_token
+        if config[:aws_secret_access_key] == iam_creds[:aws_secret_access_key] \
+          && config[:aws_access_key_id] == iam_creds[:aws_access_key_id]
+          iam_creds[:aws_session_token]
+        else
+          ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN']
+        end
       end
 
       private
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -41,14 +41,16 @@ module Kitchen
       default_config :iam_profile_name,   nil
       default_config :price,   nil
       default_config :aws_access_key_id do |driver|
-        ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] || driver.iam_creds[:aws_access_key_id]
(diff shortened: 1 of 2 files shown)


JamesAwesome:
Using ruby-1.9.3-p551 and a fresh gemset I cannot reproduce this travis build error, my local repo is in sync with the PR's code as well. 

What's stranger is one of the two tests I added passed but the error from travis was `undefined local variable or method 'iam_creds'`... If this was true both tests should have failed not just one. I'd suggest retrying the build before debugging further...

JamesAwesome:
I see what I did here... need to make these into class methods. Not sure why Rake doesn't turn that up locally.

JamesAwesome:
@tyler-ball Finally got these tests to pass. Sorry for all the chatter.

Igorshp:
@JamesAwesome there is already a 'iam_profile_name' param that assigns the role to the new node.

I think 'use_aim_profile' may be little misleading next to it.

Would be nice to do the detection automaticaly.

JamesAwesome:
@Igorshp we could change the name to something like :use_iam_role however if we try to autodetect while on an ec2 node we will always receive some form of value for `iam_creds[:aws_session_token]`, whether or not that value is valid.

Igorshp:
I mean the autodetection should check if key_id and/or secret are manualy set and return session_token accordingly.

JamesAwesome:
Might be able to whip that up, I'll futz with this a bit more and update

lib/kitchen/driver/ec2.rb
```diff
@@ -98,10 +99,10 @@ class Ec2 < Kitchen::Driver::SSHBase
         end
       end
 
-      def self.iam_creds
+      def iam_creds
         @iam_creds ||= begin
-          fetch_credentials(use_iam_profile:true)
-        rescue RuntimeError => e
+          config[:use_iam_profile] ? fetch_credentials(use_iam_profile: true) : {}
+        rescue RuntimeError, NoMethodError => e
```
tyler-ball: Where do you expect the `NoMethodError` to come from?

tyler-ball:
Thanks for adding a test and turning `fetch_credentials` into an instance method rather than class method.  I think it makes more sense for our use case and I didn't think of that.

I'm not familiar enough with AWS - is there a scenario where you would ever set access_key_id/secret_access_key AND session_token?  Or are they mutually exclusive forms of authentication?

If you can't use both of them in 1 auth call I agree with @Igorshp.  We should detect the presence of access_key_id/secret_access_key and only provide those if available.  If those are not available, use the session_token.  In this sense, the access_key (if available) overrides the session_token.  I think this makes sense because you can simply not provide those and rely on `fetch_credentials` if you want to use the session_token.

I would love to get @fnichol's opinion on this.

Igorshp:
@tyler-ball the core of the problem is the fact that you can only use one set of credentials that comes from one source.

i.e. either all are from iam_creds or all are user defined (session token can be user provided, but i've never seen it used like that. default is nil), because they are not compatible with each other.

so if even one piece of credentials is manually set, the rest should not be pulled from iam_role.

JamesAwesome:
@tyler-ball whoops I regressed the `include` to `extend` at some point. Corrected it. There are in fact instances where session token makes sense however they seem very edge. See docs [here](https://docs.aws.amazon.com/STS/latest/UsingSTS/Welcome.html).

Auto detection should work fine now.

Igorshp:
:+1:  sweet, thanks!

lib/kitchen/driver/ec2.rb
JamesAwesome: fetch_credentials calls super when it fails, so since we don't have a super for it to fail over to we need to catch NoMethodError, I think I made a note in the commit message.

JamesAwesome:
Ready for review. No more commits for now.

moimikey:
:shipit:

lib/kitchen/driver/ec2.rb
```diff
@@ -42,13 +41,14 @@ class Ec2 < Kitchen::Driver::SSHBase
       default_config :iam_profile_name,   nil
       default_config :price,   nil
       default_config :aws_access_key_id do |driver|
-        ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] || iam_creds[:aws_access_key_id]
+        ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] || driver.iam_creds[:aws_access_key_id]
       end
       default_config :aws_secret_access_key do |driver|
-        ENV['AWS_SECRET_KEY'] || ENV['AWS_SECRET_ACCESS_KEY'] || iam_creds[:aws_secret_access_key]
+        ENV['AWS_SECRET_KEY'] || ENV['AWS_SECRET_ACCESS_KEY'] \
```
tyler-ball: If you want to break this across multiple lines you don't need the `\` - just leave the `||` at the end of the line

lib/kitchen/driver/ec2.rb
```diff
@@ -98,10 +98,10 @@ class Ec2 < Kitchen::Driver::SSHBase
         end
       end
 
-      def self.iam_creds
+      def iam_creds
         @iam_creds ||= begin
-          fetch_credentials(use_iam_profile:true)
-        rescue RuntimeError => e
+          fetch_credentials(use_iam_profile: true)
+        rescue RuntimeError, NoMethodError => e
```
tyler-ball: I still have the same question about `NoMethodError` - why is it needed?  After including the ServiceMethods, you shouldn't ever encounter a `NoMethodError`.  Or am I missing something?

JamesAwesome: https://github.com/fog/fog-aws/blob/master/lib/fog/aws/credential_fetcher.rb#L27

fetch_credentials will always call super if it fails

tyler-ball: NM, I was able to reproduce the error from https://github.com/test-kitchen/kitchen-ec2/pull/104/files locally and I see why you have the `NoMethodError` check

lib/kitchen/driver/ec2.rb
```diff
@@ -167,6 +167,15 @@ def default_public_ip_association
         !!config[:subnet_id]
       end
 
+      def default_aws_session_token
+        if config[:aws_secret_access_key] == iam_creds[:aws_secret_access_key] \
+          && config[:aws_access_key_id] == iam_creds[:aws_access_key_id]
+          iam_creds[:aws_session_token]
+        else
```
tyler-ball: Why is the user required to specify the same access key / secret as returned by the iam_creds query?  If they have to do that, why not just use the provided access key / secret for authentication?

I thought the primary point of the `iam_creds` was they allowed the EC2 node acting as a provisioner to provision nodes using the local credentials.

JamesAwesome: Actually I'm just checking that config[:aws_secret_access_key] et all had fallen back to the iam_creds statement so that we know it would be logical to return the iam_creds[:aws_session_token] since our other config values are from iam.

tyler-ball:
When running on a non-ec2 node it takes a really long time to figure out the metadata service isn't available.  With that in mind I created the following patch:

```patch
diff --git a/lib/kitchen/driver/ec2.rb b/lib/kitchen/driver/ec2.rb
index e8df7af..de5eaf9 100644
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -41,14 +41,16 @@ module Kitchen
       default_config :iam_profile_name,   nil
       default_config :price,   nil
       default_config :aws_access_key_id do |driver|
-        ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] || driver.iam_creds[:aws_access_key_id]
+        ENV['AWS_ACCESS_KEY'] || ENV['AWS_ACCESS_KEY_ID'] ||
+          driver.iam_creds[:aws_access_key_id]
       end
       default_config :aws_secret_access_key do |driver|
         ENV['AWS_SECRET_KEY'] || ENV['AWS_SECRET_ACCESS_KEY'] ||
           driver.iam_creds[:aws_secret_access_key]
       end
       default_config :aws_session_token do |driver|
-        driver.default_aws_session_token
+        ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN'] ||
+          driver.iam_creds[:aws_session_token]
       end
       default_config :aws_ssh_key_id do |driver|
         ENV['AWS_SSH_KEY_ID']
@@ -98,10 +100,18 @@ module Kitchen
         end
       end

+      # First we check the existence of the metadata host.  Only fetch_credentials
+      # if we can find the host.  Ping logic taken from
+      # http://stackoverflow.com/questions/7519159/ruby-ping-pingecho-missing
       def iam_creds
+        require 'net/http'
+        require 'timeout'
         @iam_creds ||= begin
+          timeout(5) do
+            Net::HTTP.get(URI.parse('http://169.254.169.254'))
+          end
           fetch_credentials(use_iam_profile: true)
-        rescue RuntimeError, NoMethodError => e
+        rescue Errno::EHOSTUNREACH, Timeout::Error, StandardError => e
           debug("fetch_credentials failed with exception #{e.message}:#{e.backtrace.join("\n")}")
           {}
         end
@@ -167,15 +177,6 @@ module Kitchen
         !!config[:subnet_id]
       end

-      def default_aws_session_token
-        if config[:aws_secret_access_key] == iam_creds[:aws_secret_access_key] \
-          && config[:aws_access_key_id] == iam_creds[:aws_access_key_id]
-          iam_creds[:aws_session_token]
-        else
-          ENV['AWS_SESSION_TOKEN'] || ENV['AWS_TOKEN']
-        end
-      end
-
       private

       def connection
```

I think thats a better solution for this case.  If we can access the metadata service, we do.

JamesAwesome:
My only concern is that `http://169.254.169.254` is a generic link-local address and also used to serve metadata for clouds like [openstack](http://docs.openstack.org/admin-guide-cloud/content/section_metadata-service.html).

We could work around this by letting `fetch_credentials` fail and rescue NoMethodError but there may be a better way to check. Will take a look.

tyler-ball:
@JamesAwesome Ah, I see your concern - drat that people are starting to use Test Kitchen in their CI workflow from multiple cloud providers! :smile: 

But even if we can connect to `http://169.254.169.254`, `fetch_credentials` should fail because it is requesting the path `/latest/meta-data/iam/security-credentials/`.  In this case we will fall through to the `NoMethodError` rescue.

My concern is that I don't want to wait a full minute for all ec2 creates like the person in https://github.com/test-kitchen/kitchen-ec2/pull/104 is experiencing.

JamesAwesome:
Yeah I'm perfectly happy with that then if you are. :+1: I'll merge in the patch and get the tests adjusted.

JamesAwesome:
So problem:

Because this patch will call `driver.iam_creds[:aws_session_token]` if `config[:aws_session_token]` is unset in .kitchen.yml or not present in the environment we are back in the position where if you have explicitly set :aws_access_key_id or :aws_secret_access_key and are running on an ec2 host you will have an unwanted and incorrect :aws_session_token and authentication will fail.

We can deal with this by comparing the configured :aws_access_key_id against the value returned by iam_creds like I do in the current commit, however I'd like to hear @tyler-ball 's opinion on that before I continue so that we can solve this a bit quicker. I'll think on the topic a bit and comment if I come up with anything a bit more clever.

JamesAwesome:
Updated (failing) tests illustrate the problem

JamesAwesome:
@tyler-ball I lied about waiting for a response. I think this is the best I can do. All unprevented combinations of autodetectable credentials via iam_creds and user defined settings should simply fail authentication...

The `required_config` method made testing default values being set from `iam_creds` rather difficult. I chose to remove them since the ENV defaults for those values are untested as well (presumably for the same reason). I could add them back if we remove the `required_config` statements for :aws_access_key_id and :aws_secret_access_key but that would probably be undesired.

If you know of a way to get those tests to pass without having to remove `required_config` i'd be happy to implement them as well.

JamesAwesome:
I am now 100% happy with this.

I had to remove the top level `before` block since it was calling `instance` too soon which was causing my tests to fail. 

@tyler-ball I promise no more commits unless you comment. :smile:

tyler-ball:
@JamesAwesome I did a little bit of test refactoring / cleanup when I merged, but thank you for tackling this issue and doing all the work!

JamesAwesome:
:+1: :smile:

Igorshp:
:+1:  Thanks

PR #109, opened 2015-04-18
Updating to depend on TK 1.4

tyler-ball:
Taking advantage of the new base class and transport logic.  Also replacing the Fog dependency with the aws-sdk dependency.

\cc @fnichol


lib/kitchen/driver/ec2.rb
```diff
@@ -99,65 +131,83 @@ class Ec2 < Kitchen::Driver::SSHBase
...
+      def finalize_config!(instance)
+        super
+        if config[:availability_zone].nil?
+          config[:availability_zone] = config[:region]+'b'
+        end
+        if config[:instance_type].nil?
+          config[:instance_type] = config[:flavor_id] || 'm1.small'
+        end
+        if config[:ssh_timeout]
+          # TODO NOOOOOOOOOO
```
tyler-ball: @fnichol I really, really hate doing this.  What would you suggest?

kitchen-ec2.gemspec
```diff
@@ -18,8 +18,11 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.require_paths = ["lib"]
 
-  gem.add_dependency 'test-kitchen', '~> 1.0'
-  gem.add_dependency 'fog'
+  gem.add_dependency "test-kitchen", "= 1.4.0.rc.1"
+  gem.add_dependency 'excon'
+  gem.add_dependency 'multi_json'
```
fnichol: Is there a reason for multi_json other than older Ruby support? Test Kitchen only works on Ruby 1.9 and up which ships with a 'json' in stdlib. I may be missing a hidden dep in here though…

lib/kitchen/driver/ec2.rb
```diff
@@ -28,18 +33,25 @@ module Driver
     # Amazon EC2 driver for Test Kitchen.
     #
     # @author Fletcher Nichol <fnichol@nichol.ca>
-    class Ec2 < Kitchen::Driver::SSHBase
-      include Fog::AWS::CredentialFetcher::ServiceMethods
+    class Ec2 < Kitchen::Driver::Base
+
+      kitchen_driver_api_version 2
```
fnichol: Nice :smile:

lib/kitchen/driver/ec2.rb
```diff
@@ -194,84 +247,176 @@ def default_aws_session_token
...
+        end
+        if config[:ssh_retries]
+          state[:connection_retries] = config[:ssh_retries]
+        end
+        if config[:username]
+          state[:username] = config[:username]
+        elsif instance.transport[:username] == instance.transport.class.defaults[:username]
+          # If the transport has the default username, copy it from amis.json
+          # This duplicated old behavior but I hate amis.json
+          ami_username = amis['usernames'][instance.platform.name]
```
fnichol: We'll have to work this out longer term--ideally the Driver should be able to pick an appropriate default username (based on AMI assuming we have that info), but until then the behavior is maintained which is good.

lib/kitchen/driver/ec2.rb
```diff
@@ -57,107 +69,152 @@ class Ec2 < Kitchen::Driver::SSHBase
...
-      default_config :ssh_retries, 3
 
       required_config :aws_access_key_id
       required_config :aws_secret_access_key
       required_config :aws_ssh_key_id
       required_config :image_id
 
+      def self.validation_warn(driver, old, new)
+        driver.warn "WARN: The driver[#{driver.class.name}] config key `#{old}` is deprecated," +
+          " please use `#{new}`"
```
fnichol: I honestly don't know, but could `new` possibly be construed as `Object.new`? Maybe `old_key`/`new_key` despite the names being longer?

lib/kitchen/driver/ec2.rb
```diff
@@ -57,107 +69,152 @@ class Ec2 < Kitchen::Driver::SSHBase
...
       def create(state)
+        copy_deprecated_configs(state)
         return if state[:server_id]
 
-        info("Creating <#{state[:server_id]}>...")
-        info('If you are not using an account that qualifies under the AWS')
-        info('free-tier, you may be charged to run these suites. The charge')
-        info('should be minimal, but neither Test Kitchen nor its maintainers')
-        info('are responsible for your incurred costs.')
+        info(<<-END.gsub!(/^\s+/m,''))
```
fnichol: There is a `Kitchen::Util.outdent!` that could help here:

```ruby
info(Kitchen::Util.oudent!(<<-END))
```

lib/kitchen/driver/ec2.rb
```diff
@@ -57,107 +69,152 @@ class Ec2 < Kitchen::Driver::SSHBase
...
-        print '(server ready)'
+
+        info("EC2 instance <#{state[:server_id]}> ready.")
         state[:hostname] = hostname(server)
-        wait_for_sshd(state[:hostname], config[:username], {
-          :ssh_timeout => config[:ssh_timeout],
-          :ssh_retries => config[:ssh_retries]
-        })
-        print "(ssh ready)\n"
+        instance.transport.connection(state).wait_until_ready
```
fnichol: Nice! This is the one-liner I'm hoping will resolve a ton of connection issues with the Driver

lib/kitchen/driver/ec2.rb
```diff
@@ -194,84 +247,176 @@ def default_aws_session_token
         env
       end
 
+      # First we check the existence of the metadata host.  Only fetch_credentials
+      # if we can find the host.
+      def iam_creds
+        require 'net/http'
+        require 'timeout'
```
fnichol: You could push these requires up  to the start of the .rb file, should be okay

lib/kitchen/driver/ec2.rb
```diff
@@ -194,84 +247,176 @@ def default_aws_session_token
...
+          {}
+        end
+      end
+
+      INSTANCE_METADATA_HOST = 'http://169.254.169.254'
+      INSTANCE_METADATA_PATH = '/latest/meta-data/iam/security-credentials/'
+      # fetch_credentials logic copied from Fog
+      def fetch_credentials
+        begin
+          connection = Excon.new(INSTANCE_METADATA_HOST)
```
fnichol: Might be able to get away with vanilla `net/http` which could trim excon as a gem dependency, but we could also look at that later in another release

lib/kitchen/driver/ec2.rb
```diff
@@ -194,84 +247,176 @@ def default_aws_session_token
...
-        debug("ec2:user_data '#{config[:user_data]}'")
-        debug("ec2:ssh_timeout '#{config[:ssh_timeout]}'")
-        debug("ec2:ssh_retries '#{config[:ssh_retries]}'")
-        debug("ec2:spot_price '#{config[:price]}'")
+        debug('EC2 Server Configuration')
+        names = [
+          :region, :availability_zone, :instance_type, :ebs_optimized, :image_id,
+          :private_ip_address, :security_group_ids, :tags, :aws_ssh_key_id, :subnet_id,
+          :iam_profile_name, :associate_public_ip, :user_data, :price
+        ]
```
fnichol: That seems much more scalable in the long term :smile: 

Ultimately, I wonder how much this is used and if users would get similar benefit from running `kitchen diagnose` (the seed of this code pattern pre-dated the diagnose command). Anway, topic for another day--this is much better.

lib/kitchen/driver/ec2.rb
```diff
@@ -320,68 +458,69 @@ def self.hostname(server, interface_type=nil)
...
+          if config[:ebs_volume_size] || config[:ebs_delete_on_termination] ||
+            config[:ebs_device_name] || config[:ebs_volume_type]
+            # If the user didn't supply block_device_mappings but did supply
+            # the old configs, copy them into the block_device_mappings array correctly
+            bdm = {}
+            if bdm[:ebs_volume_size].nil?
+              bdm[:ebs_volume_size] = config[:ebs_volume_size] || 8
+            end
+            if bdm[:ebs_delete_on_termination].nil?
+              bdm[:ebs_delete_on_termination] = config[:ebs_delete_on_termination] || true
```
fnichol: Could `config[:ebs_delete_on_termination]` be set to `false`? If so using something like `config.fetch(:ebs_delete_on_termination, true)` may end up being safer (which is supported by `Kitchen::LazyHash`).

fnichol:
@tyler-ball Looking really good!

kitchen-ec2.gemspec
tyler-ball: When getting rid of Fog, I literally copied the transitive dependencies it was using (for the `fetch_credentials` call).  I should probably update that to just use json.

lib/kitchen/driver/ec2.rb
tyler-ball: You're right!  I made the same comment on one of Dan's reviews.  Thanks!

lib/kitchen/driver/ec2.rb
tyler-ball: gracias!

lib/kitchen/driver/ec2.rb
tyler-ball: @fnichol Yes, `config[:ebs_delete_on_termination]` could be false.  Thanks for the catch!  I'll update to use fetch.

lib/kitchen/driver/ec2.rb
tyler-ball: I agree - kitchen diagnose is a much better pattern, but as you have said, lets take it in steps.

fnichol:
Sweet, :+1:

Issue #66, opened 2014-09-03
Dead link in kitchen-ec2/README


luntzel:
Link to http://docs.kitchen-ci.org/drivers/usage at:

Please read the "Driver usage" page for more details.

is toast :(

Joseph-R:
+1.  Just ran into this myself.

http://docs.kitchen-ci.org/drivers/usage

proffalken:
@luntzel wow, it's been a whilst since this was logged!

Please can you submit a PR with the appropriate changes and link it back to this issue? Once we've got that I'll talk to @cheeseplus and try to get it merged ASAP assuming it's still relevant.

If it's not relevant, please close this issue.

PR #119, opened 2015-05-19
Issue #66  - changed [driver_usage] link to point to Rubygems.org kitchen-ec2 gem

dsavinkov:


tyler-ball:
I don't think the Rubygems link provides any more useful information, so I think this link line should just get replaced with the reference to the github page.  Can you make that update?

dsavinkov:
@tyler-ball - Done in #141

Issue #122, opened 2015-05-20
hostname missing when waiting for ssh service in create action 


litjoco:
During the `create` action the instance is created and logged as `ready`.  Following this the ssh connection is being checked.  The connection is always fails made and eventually times out. I am able to manually connect during this time using same keypair.  From the log it looks like the hostname is not getting set.  

I have tried explicitly setting the `interface_type` and leaving it as default with same result.

```
       Waiting for SSH service on :22, retrying in 3 seconds
D      [SSH] opening connection to ec2-user@<{:user_known_hosts_file=>"/dev/null", :paranoid=>false, :port=>22, :compression=>"zlib", :compression_level=>6, :keepalive=>true, :keepalive_interval=>60, :timeout=>120, :keys_only=>true, :keys=>["/some/path/key.pem"], :user=>"ec2-user"}>
D      [SSH] connection failed (#<Errno::ECONNREFUSED: Connection refused - connect(2) for "" port 22>)
```

fyi: After the time out the state file is created with the instance id but hostname is empty

dsavinkov:
@litjoco 
Make sure to adjust your `.kitchen.yml` file:
```yml
...
platforms:
  - name: centos-6.3
    driver_plugin: ec2
    transport:
      username: test
      ssh_key: <%= ENV['AWS_SSH_KEY']%>
...
```
It works correctly for me.

callmeradical:
If the hostname is not set, we are having an issue where it attempts to converge on localhost. This is very bad.

huntertj:
I'm having the same problem, I noticed the .kitchen/default-<name> is missing the hostname field, no matter if I specify the lines suggested by @litjoco, also tried specifying interface: to dns, internal or public with the same results
```bash
last_action: destroy
server_id: i-5ec8f688
hostname: ''
```

litjoco:
Thanks for the suggestion @dsavinkov.
If I manually update the state file to have the hostname or ip, that should be  populated, then the additional actions such as `converge`, `verify` and `login` all work as expected.  

This makes me think that my config is correct

callmeradical:
I have a merge request I am about to file to default to returning the private ip address

PR #123, opened 2015-05-20
testing 0.9.0 fix

callmeradical:
This wires up the interface variable in the config and defaults to returning the private_ip_address for the hostname.

PR #124, opened 2015-05-20
Fixing error where aws returns DNS name as empty string

tyler-ball:
Fixes https://github.com/test-kitchen/kitchen-ec2/issues/122
Replaces https://github.com/test-kitchen/kitchen-ec2/pull/123

\cc @fnichol @litjoco @dsavinkov @callmeradical @huntertj

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -20,9 +20,10 @@ require "benchmark"
 require "json"
 require "aws"
 require "kitchen"
-require "kitchen/driver/ec2_version"
+require_relative "ec2_version"
 require_relative "aws/client"
 require_relative "aws/instance_generator"
+require "aws-sdk-core/waiters/errors"
 
 module Kitchen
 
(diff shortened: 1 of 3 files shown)


tyler-ball:
It looks like I also broke backwards compatibility when I removed the `interface` config - adding back in this PR as well!

lib/kitchen/driver/ec2.rb
```diff
@@ -354,6 +361,8 @@ def hostname(server, interface_type = nil)
           potential_hostname = nil
           INTERFACE_TYPES.values.each do |type|
             potential_hostname ||= server.send(type)
+            # AWS returns an empty string if the dns name isn't populated yet
```
callmeradical: Why not default to private_ip unless specified? 

It seems like it would be the most likely to be available, followed by public_ip and then DNS. Looks like there could be reduced complexity unless there is something I am missing. 

What is the behavior of not defining the hostname? (returning nil)

lib/kitchen/driver/ec2.rb
```diff
@@ -194,14 +196,21 @@ def create(state) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
...
-          hostname = hostname(s)
-          # Euca instances often report ready before they have an IP
-          s.state.name == "running" && !hostname.nil? && hostname != "0.0.0.0"
+        begin
+          server = server.wait_until(
+            :max_attempts => config[:retryable_tries],
+            :delay => config[:retryable_sleep],
+            :before_attempt => wait_log
+          ) do |s|
+            hostname = hostname(s, config[:interface])
```
huntertj: Only problem I can see with this is if a user has interface set in their .kitchen.yml to public and associate_public_ip = false then kitchen will never detect its running

lib/kitchen/driver/ec2.rb
tyler-ball: `private_ip` is the most likely to be available but not the most likely to be reachable by SSH.  The logic here is currently

```ruby
# if `interface_type` is provided:
  # query only that interface_type
# else
  # first, try dns
  # second, try public_ip
  # third, try private_ip
  # return the first value that is non-nil and non-empty
```

This brings the `hostname` logic in line with what is in the [README](https://github.com/test-kitchen/kitchen-ec2/blob/aa765caf608efd07040a527cf2cfcb05daa3f87b/README.md#interface).  I don't want to break the backwards compatibility with logic that has been there for a while.

I also like the 'try from easiest to SSH into to hardest to SSH into' logic rather than 'try from what is most likely to be available to what is least likely to be available'.  The goal is to get a hostname we can SSH into and then install Chef.  If a user really wants to use private_ip they can specify `interface: private` in their .kitchen.yml.

lib/kitchen/driver/ec2.rb
tyler-ball: That is true, but I would call that user error - the user has told Test Kitchen to query a parameter that is never going to be available.  A `kitchen diagnose` should reveal this.

lib/kitchen/driver/ec2.rb
callmeradical: +1 I was thinking of this soley from inside a VPC, good call. Little short-sighted on my part.

fnichol:
LGTM, :+1:

Issue #126, opened 2015-05-21
Support for proxy?


litjoco:
I am seeing failures when behind a proxy which was working on a previous version. I have the proxy environment variables set. 

I think the ability to connect from behind a proxy has been lost when moving from fog to the AWS sdk.

I'm not familiar with the aws sdk, but think the ec2 client [can be supplied a http_proxy in the config option](http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html).

tyler-ball:
@litjoco how are you providing the proxy?  Through the .kitchen.yml?

litjoco:
@tyler-ball Yes- the environment variables are set.  Having these set worked previously-with a commit ref pre-aws sdk I believe. I have the aws cli installed and it works as expected with the env vars set.

Thanks for your work on helping resolve

tyler-ball:
@litjoco I added a comment in the PR, but I'll ask here too: I'm currently defaulting the http_proxy to `ENV['HTTP_PROXY']` but because the AWS SDK V2 only accepts 1 proxy would you rather it defaulted to `ENV['HTTPS_PROXY'] || ENV['HTTP_PROXY']`?

metadave:
looks good, +1 for `ENV['HTTPS_PROXY'] || ENV['HTTP_PROXY']`

litjoco:
@tyler-ball I added a comment to the pull request in agreement with @metadave 

thank you

PR #131, opened 2015-05-26
Adding proxy support that was present in Fog back

tyler-ball:
Fixes https://github.com/test-kitchen/kitchen-ec2/issues/126

\cc @fnichol @litjoco

I added a config option that defaults to `ENV['HTTP_PROXY']`.  @litjoco as you pointed out, there is only 1 `:http_proxy` setting available in the AWS Ruby SDK V2, so we cannot specify a different `:https_proxy`.  But I set up a local [mitmproxy](https://mitmproxy.org/doc/mitmproxy.html) and tested that HTTPS requests worked through this proxy when specifying `ENV['HTTP_PROXY']`

But because all the requests to AWS were actually https requests, I'm thinking of trying to first default to `ENV['HTTPS_PROXY']` then falling back to `ENV['HTTP_PROXY']`.  Would this support your use case better?  Or are both environment variables typically pointing to the same proxy?

I'm also adding in the 0.9.2 release prep here because this is the last PR I'm waiting on for that release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+## 0.9.2 / 2015-05-26
+
+### Bug Fixes
+
+* Pull Request [#131][]: Adding back support for a proxy via config `http_proxy` which defaults to `ENV['HTTP_PROXY']`
+* Pull Request [#128][]: 
+    * Fixes [#121][]: Fixing error `Network interfaces and an instance-level security groups may not be specified on the same request` 
+    * Fixes [#127][]: User Data content should be base64 encoded when passed to aws sdk 
+
+### New Features
+
+### Improvements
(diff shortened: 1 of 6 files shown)


litjoco:
@tyler-ball yes - I agree with your thinking - prefer `HTTPS_PROXY` if set

tyler-ball:
@metadave @litjoco I updated to prefer HTTPS over HTTP - as soon as Travis finishes I'll release 0.9.2!

fnichol:
That's looking good to me. Good idea preserving the `:http_proxy` config name though as it's used in other plugins, so easier consume that.

fnichol:
Oh, and :+1: , heh

PR #133, opened 2015-05-28
Add feature to get Windows password generated by AWS

zl4bv:
Rather than pre-determining the password for a Windows instance (baked AMI, set password in user-data, etc) it would be nice to be able to use the password generated automatically by AWS.

Example config:
```yaml
---
driver:
  name: ec2
  aws_ssh_key_id: key_pair_name
  get_windows_password: true
 
transport:
  name: winrm
  username: Administrator
  ssh_key: /path/to/key_pair_name.pem
```


tyler-ball:
Rather than adding a new config key, shouldn't we just fetch the password if they have supplied `config[:username] =~ /[Aa]dministrator/` and haven't supplied `config[:password]`?

zl4bv:
Yeah that makes more sense.

My concern is when this is used with a Linux instance with an SSH key (therefore not specifying a password). If we're happy to assume that someone running kitchen-ec2 with a Linux instance won't use a username matching `/[Aa]dministrator/` then it should be fine.

tyler-ball:
The driver has a `windows_os?` method that can tell if the instance its spinning up is a Windows instance

tyler-ball:
Closing this as it was merged as part of https://github.com/test-kitchen/kitchen-ec2/pull/150

PR #136, opened 2015-05-28
block device examples updated

gmiranda23:
The block device examples are incorrect.  Adjusted them to reflect a useable config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,10 +232,10 @@ name of your ebs device, for example: `/dev/sda1`
 A list of block device mappings for the machine.  An example of all available keys looks like:
 ```yaml
 block_device_mappings:
-  - ebs_device_name: /dev/sda1
+  - ebs_device_name: /dev/sda
     ebs_volume_size: 20
     ebs_delete_on_termination: true
-  - ebs_device_name: /dev/sda2
+  - ebs_device_name: /dev/sdb
     ebs_volume_type: gp2
     ebs_virtual_name: test
     ebs_volume_size: 15
@@ -371,7 +371,7 @@ driver:
   private_ip_address: 10.0.0.27
   interface: dns
   block_device_mappings:
-    - ebs_device_name: /dev/sda0
+    - ebs_device_name: /dev/sda
       ebs_volume_type: gp2
       ebs_virtual_name: test
       ebs_volume_size: 15
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ driver:
   aws_ssh_key_id: id_rsa-aws
   security_group_ids: ["sg-1a2b3c4d"]
   region: us-west-2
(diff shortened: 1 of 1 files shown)


tyler-ball:
Hey @gmiranda23 - I actually get errors trying to use this PR.  What is the error you're seeing?

Although, while troubleshooting I did come across an unexpected exception trying to call `logger` from the InstanceGenerator class

gmiranda23:
It's not so much an error as misleading info.  I don't typically come across EBS volumes with block device trailing numbers for anything other than root volumes.  And even then, most newer instance types have problems with them.  I wasn't quite sure why, but your question prompted me to look it up.

I got some clarification on this from the [EBS Block Device Mapping Concepts](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html) doc.  

"Trailing numbers on device names can sometimes be used, but may cause conflicts. [...]  Hardware virtual machine (HVM) AMIs don't support the use of trailing numbers on device names (xvd[a-p][1-15])."

So perhaps to make that usage point, the example should have one definition with a trailing device number and one without?

tyler-ball:
If it is suggested to use non-numbered device names, then I'm all for updating the documentation.  I'm not very familiar with the concept of disc device naming.

I would just like to make sure the example does work.  Can you update the AMIs in the example at the bottom of the README so specifying `/dev/sda` instead of `/dev/sda1` will create it correctly?  Perhaps just use Ubuntu 15.04 HVM (`ami-cf09e8a4`) and Centos 7 HVM (`ami-96a818fe`).  Will you try those AMIs and test the full example to make sure it works?

gmiranda23:
After some research, I put together a full working .kitchen.yml example using non-standard settings (us-west-2), with updated fields where appropriate, a mix of using platforms defined in amis.json, and custom platforms with different settings.  Makes usage described in #153 clear.

README.md
```diff
@@ -361,21 +361,14 @@ driver:
   name: ec2
   aws_ssh_key_id: id_rsa-aws
   security_group_ids: ["sg-1a2b3c4d"]
-  region: us-east-1
-  availability_zone: b
+  region: us-west-2
+  availability_zone: us-west-2b
```
tyler-ball: I know this was broken earlier, but I would like to encourage people to use the single-letter form of this config to DRY this yaml up a bit.  Would you try `b` with the latest version of kitchen-ec2 (0.9.5+)?

tyler-ball:
Other than my last minor comment, this is looking awesome.  Thanks @gmiranda23!

gmiranda23:
I updated it to the long format because single letter was busted.  I agree that's a better form.  Tested on kitchen-ec2 0.9.5 and it works again.  LGTM!

PR #139, opened 2015-05-29
Delay tagging server until after instance is ready.

fnichol:
This helps to ensure that tagging isn't attempted before the instance
can be referenced before it exists.


fnichol:
Closing in favor of #140 which has a nicer solution.

Issue #138, opened 2015-05-29
Error trying to tag instance before it exists


tyler-ball:
```
I, [2015-05-28T20:26:05.427218 #22655]  INFO -- Kitchen: -----> Starting Kitchen (v1.4.0)
I, [2015-05-28T20:26:05.856924 #22655]  INFO -- Kitchen: -----> Cleaning up any prior instances of <webapp-ubuntu-1204>
I, [2015-05-28T20:26:05.857180 #22655]  INFO -- Kitchen: -----> Destroying <webapp-ubuntu-1204>...
I, [2015-05-28T20:26:05.858304 #22655]  INFO -- Kitchen: -----> Testing <webapp-ubuntu-1204>
I, [2015-05-28T20:26:05.858469 #22655]  INFO -- Kitchen: -----> Creating <webapp-ubuntu-1204>...
E, [2015-05-28T20:26:07.249275 #22655] ERROR -- Kitchen: ------Exception-------
E, [2015-05-28T20:26:07.249342 #22655] ERROR -- Kitchen: Class: Kitchen::ActionFailed
E, [2015-05-28T20:26:07.249385 #22655] ERROR -- Kitchen: Message: Failed to complete #create action: [The instance ID 'i-e55fc612' does not exist]
E, [2015-05-28T20:26:07.249426 #22655] ERROR -- Kitchen: ---Nested Exception---
E, [2015-05-28T20:26:07.249476 #22655] ERROR -- Kitchen: Class: Aws::EC2::Errors::InvalidInstanceIDNotFound
E, [2015-05-28T20:26:07.249516 #22655] ERROR -- Kitchen: Message: The instance ID 'i-e55fc612' does not exist
E, [2015-05-28T20:26:07.249554 #22655] ERROR -- Kitchen: ------Backtrace-------
E, [2015-05-28T20:26:07.249595 #22655] ERROR -- Kitchen: /home/travis/.rvm/gems/ruby-2.1.5/gems/aws-sdk-core-2.0.46/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'
E, [2015-05-28T20:26:07.249640 #22655] ERROR -- Kitchen: /home/travis/.rvm/gems/ruby-2.1.5/gems/aws-sdk-core-2.0.46/lib/seahorse/client/plugins/param_conversion.rb:22:in `call'
[...]

PR #140, opened 2015-05-29
Adding an existence check before tagging server

tyler-ball:
fixes https://github.com/test-kitchen/kitchen-ec2/issues/138

\cc @fnichol @danielsdeleo

### DIFF
--- a/.cane
+++ b/.cane
@@ -1,2 +1,2 @@
---abc-max 30
+--abc-max 40
 --style-measure 100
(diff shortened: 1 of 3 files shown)


danielsdeleo:
Not too familiar with the codebase, but it looks sensible to me based on our understanding of the issue. :+1:

fnichol:
Ha, I started to attack this in #139, however adding that wait is a much better solution IMO. I'll close mine out in favor of this and give a :+1:

PR #134, opened 2015-05-28
Make Ec2#create_ec2_json work on Windows guests

zl4bv:
The Kitchen::Driver::Ec2#create_ec2_json method works on Linux and OS X but not on Windows.

This PR fixes the following error when the method runs against a Windows guest:

```
       Waiting for WinRM service on http://ec2-52-64-76-82.ap-southeast-2.compute.amazonaws.com:5985/wsman, retrying in 3 seconds
       [WinRM] Established
$$$$$$ sudo : The term 'sudo' is not recognized as the name of a cmdlet, function,
$$$$$$ script file, or operable program. Check the spelling of the name, or if a path
$$$$$$ was included, verify that the path is correct and try again.
$$$$$$ At line:1 char:1
$$$$$$ + sudo mkdir -p /etc/chef/ohai/hints;sudo touch /etc/chef/ohai/hints/ec2.json
$$$$$$ + ~~~~
$$$$$$     + CategoryInfo          : ObjectNotFound: (sudo:String) [], CommandNotFoun
$$$$$$    dException
$$$$$$     + FullyQualifiedErrorId : CommandNotFoundException
$$$$$$
$$$$$$ sudo : The term 'sudo' is not recognized as the name of a cmdlet, function,
$$$$$$ script file, or operable program. Check the spelling of the name, or if a path
$$$$$$ was included, verify that the path is correct and try again.
$$$$$$ At line:1 char:36
$$$$$$ + sudo mkdir -p /etc/chef/ohai/hints;sudo touch /etc/chef/ohai/hints/ec2.json
$$$$$$ +                                    ~~~~
$$$$$$     + CategoryInfo          : ObjectNotFound: (sudo:String) [], CommandNotFoun
$$$$$$    dException
$$$$$$     + FullyQualifiedErrorId : CommandNotFoundException
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: Failed to complete #create action: [WinRM exited (1) for command: [sudo mkdir -p /etc/chef/ohai/hints;sudo touch /etc/chef/ohai/hints/ec2.json]]
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

PR #96, opened 2015-02-13
Creating windows-guest-support branch for kitchen-ec2

tyler-ball:
This tracks the changes in https://github.com/test-kitchen/test-kitchen/tree/windows-guest-support - when that is merged to master, this should be also.

These changes come from https://github.com/afiune/kitchen-ec2 .  I had to clean up some merge conflicts, so it will be slightly different.  Tests are also not going to pass yet - I need to clean up formatting, but it is in another branch for now since it makes the diff impossible to read.

\cc @afiune @fnichol @test-kitchen/windows 

NTS - merge afiune-transport-2 into this branch before going to master

PR #150, opened 2015-06-04
Pulling together existing PRs for windows support

tyler-ball:
Checklist of items (copied from https://github.com/test-kitchen/kitchen-ec2/pull/133)

- [x] WinRM not enabled by default on EC2 instances so have to manually enable it with user-data
- [x] This issue - have to wait for WinRM to be enabled; WinRM transport currently too impatient
- [x] Default username for WinRM transport doesn't work (issue test-kitchen/test-kitchen#688)
- [x] Command to create EC2 JSON for Ohai doesn't work because the command is targeted at *NIX-based machines - which breaks the end of the kitchen 'create' phase
- [x] WinRM enabled automatically by Kitchen-ec2 (could be salvaged from #96)
- [x] EC2-generated Administrator password (#133)
- [x] Windows AMIs in `amis.json`
- ~~Store generated admin password and pass it to RDP command when running `kitchen login`~~
- [x] Add test coverage for new methods and `create` method
- [x] Documentation - windows support, specifying username & password, auto generated user_data

Note: The only way I could see to store a password in the `.rdp` file is to encrypt it using a microsoft-only DLL

Depends on https://github.com/test-kitchen/test-kitchen/pull/736
Replaces https://github.com/test-kitchen/kitchen-ec2/pull/96
Replaces https://github.com/test-kitchen/kitchen-ec2/pull/133
Includes https://github.com/test-kitchen/kitchen-ec2/pull/134

\cc @fnichol @zl4bv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,20 @@
-## 0.9.4 / 2015-06-05
+## 0.10.0 / TBD
+
+### Bug Fixes
+
+### New Features
+
+* Pull Request [#150][]: Added support for managing Windows hosts in AWS
+    * Updated `amis.json` to include AMIs for Windows Server 2008 and 2012R2.
+    * Add default user data which enables winrm.  See `ec2.rb` for the exact user data that is uploaded.
+    * Default user data will add a non-`administrator` user to the Administrator group if `username` and `password` transport options are specified.
+    * EC2 will automatically generate the `administrator` user password if it is not specified as `password` transport options.  This will be stored and used for further kitchen commands.
+        * Unfortunately this cannot be passed to the RDP client initiated by `kitchen login`.  Look at `.kitchen/<name>.yml` for the password in this case.
+    * A HUGE thanks to [@zl4bv][] and [@afiune][] for the majority of the work on these changes!
+
+### Improvements
+
+## 0.9.5 / 2015-06-05
 
 ### Bug Fixes
 
@@ -175,6 +191,7 @@
 [#131]: https://github.com/test-kitchen/kitchen-ec2/issues/131
 [#140]: https://github.com/test-kitchen/kitchen-ec2/issues/140
 [#142]: https://github.com/test-kitchen/kitchen-ec2/issues/142
+[#150]: https://github.com/test-kitchen/kitchen-ec2/issues/150
 [#151]: https://github.com/test-kitchen/kitchen-ec2/issues/151
 [@Atalanta]: https://github.com/Atalanta
 [@Igorshp]: https://github.com/Igorshp
@@ -196,6 +213,7 @@
 [@mattray]: https://github.com/mattray
 [@mumoshu]: https://github.com/mumoshu
 [@nicgrayson]: https://github.com/nicgrayson
+[@afiune]: https://github.com/afiune
 [@scarolan]: https://github.com/scarolan
 [@sebbrandt87]: https://github.com/sebbrandt87
 [@sethvargo]: https://github.com/sethvargo
@@ -203,3 +221,4 @@
 [@spheromak]: https://github.com/spheromak
 [@tiwilliam]: https://github.com/tiwilliam
 [@tyler-ball]: https://github.com/tyler-ball
+[@zl4bv]: https://github.com/zl4bv
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -460,18 +460,14 @@ module Kitchen
         $logfile="C:\\Program Files\\Amazon\\Ec2ConfigService\\Logs\\kitchen-ec2.log"
         #PS Remoting and & winrm.cmd basic config
         Enable-PSRemoting -Force -SkipNetworkProfileCheck
-        & winrm.cmd quickconfig -q >> $logfile
         & winrm.cmd set winrm/config '@{MaxTimeoutms="1800000"}' >> $logfile
-        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="512"}' >> $logfile
+        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="1024"}' >> $logfile
         & winrm.cmd set winrm/config/winrs '@{MaxShellsPerUser="50"}' >> $logfile
-        #Client settings
-        & winrm.cmd set winrm/config/client/auth '@{Basic="true"}' >> $logfile
-        #Server settings
+        #Server settings - support username/password login
         & winrm.cmd set winrm/config/service/auth '@{Basic="true"}' >> $logfile
         & winrm.cmd set winrm/config/service '@{AllowUnencrypted="true"}' >> $logfile
         & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="1024"}' >> $logfile
         #Firewall Config
-        & netsh.exe advfirewall set publicprofile state off  >> $logfile
         & netsh advfirewall firewall set rule name="Windows Remote Management (HTTP-In)" profile=public protocol=tcp localport=5985 remoteip=localsubnet new remoteip=any  >> $logfile
         #{customAdminScript}
         </powershell>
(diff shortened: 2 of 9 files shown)


lib/kitchen/driver/ec2.rb
```diff
@@ -382,9 +425,57 @@ def hostname(server, interface_type = nil)
...
+          $seccfg = [IO.Path]::GetTempFileName()
+          & secedit.exe /export /cfg $seccfg >> $logfile
+          (Get-Content $seccfg) | Foreach-Object {$_ -replace "PasswordComplexity\\s*=\\s*1", "PasswordComplexity = 0"} | Set-Content $seccfg
+          & secedit.exe /configure /db $env:windir\\security\\new.sdb /cfg $seccfg /areas SECURITYPOLICY >> $logfile
+          & cp $seccfg "c:\\"
+          & del $seccfg
+          $username="#{instance.transport[:username]}"
+          $password="#{instance.transport[:password]}"
+          "Creating static user: $username" >> $logfile
+          & net.exe user /y /add $username $password >> $logfile
```
zl4bv: Will this silently fail if the user already exists? Thinking about baked AMIs, etc.

lib/kitchen/driver/ec2.rb
```diff
@@ -382,9 +425,57 @@ def hostname(server, interface_type = nil)
...
+        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="512"}' >> $logfile
+        & winrm.cmd set winrm/config/winrs '@{MaxShellsPerUser="50"}' >> $logfile
+        #Client settings
+        & winrm.cmd set winrm/config/client/auth '@{Basic="true"}' >> $logfile
+        #Server settings
+        & winrm.cmd set winrm/config/service/auth '@{Basic="true"}' >> $logfile
+        & winrm.cmd set winrm/config/service '@{AllowUnencrypted="true"}' >> $logfile
+        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="1024"}' >> $logfile
+        #Firewall Config
+        & netsh.exe advfirewall set publicprofile state off  >> $logfile
```
smurawski: This completely shuts off the firewall for public identified network connections.  If that's the goal the the following line is redundant.  I understand these are supposed to be short-lived instances, but if they aren't destroyed you'll have a box with no firewall open to the internet.

lib/kitchen/driver/ec2.rb
```diff
@@ -382,9 +425,57 @@ def hostname(server, interface_type = nil)
...
+        & winrm.cmd quickconfig -q >> $logfile
+        & winrm.cmd set winrm/config '@{MaxTimeoutms="1800000"}' >> $logfile
+        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="512"}' >> $logfile
+        & winrm.cmd set winrm/config/winrs '@{MaxShellsPerUser="50"}' >> $logfile
+        #Client settings
+        & winrm.cmd set winrm/config/client/auth '@{Basic="true"}' >> $logfile
+        #Server settings
+        & winrm.cmd set winrm/config/service/auth '@{Basic="true"}' >> $logfile
+        & winrm.cmd set winrm/config/service '@{AllowUnencrypted="true"}' >> $logfile
+        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="1024"}' >> $logfile
```
smurawski: Client settings are really only needed if you are initiating winrm connections from the guest computer (which gets tricky due to credential delegation in a service context).  It doesn't hurt anything, but the client settings modifications should not be necessary.

lib/kitchen/driver/ec2.rb
```diff
@@ -382,9 +425,57 @@ def hostname(server, interface_type = nil)
...
+
+        # Returning the fully constructed PowerShell script to user_data
+        Kitchen::Util.outdent!(<<-EOH)
+        <powershell>
+        $logfile="C:\\Program Files\\Amazon\\Ec2ConfigService\\Logs\\kitchen-ec2.log"
+        #PS Remoting and & winrm.cmd basic config
+        Enable-PSRemoting -Force -SkipNetworkProfileCheck
+        & winrm.cmd quickconfig -q >> $logfile
+        & winrm.cmd set winrm/config '@{MaxTimeoutms="1800000"}' >> $logfile
+        & winrm.cmd set winrm/config/winrs '@{MaxMemoryPerShellMB="512"}' >> $logfile
```
smurawski: I'd recommend 1024 for MaxMemoryPerShellMB (that was supposed to be the default for 2012 - http://stevenmurawski.com/powershell/2015/5/powershell-3-winrm-maxmemorypershellmb-is-wrong - and is the default for Server 2012 R2) .  And maybe warn if PowerShell 3 is installed, as there is a hotfix required to make WinRM use the configured memory settings.

lib/kitchen/driver/ec2.rb
```diff
@@ -382,9 +425,57 @@ def hostname(server, interface_type = nil)
...
+          EOH
+        end
+
+        # Returning the fully constructed PowerShell script to user_data
+        Kitchen::Util.outdent!(<<-EOH)
+        <powershell>
+        $logfile="C:\\Program Files\\Amazon\\Ec2ConfigService\\Logs\\kitchen-ec2.log"
+        #PS Remoting and & winrm.cmd basic config
+        Enable-PSRemoting -Force -SkipNetworkProfileCheck
+        & winrm.cmd quickconfig -q >> $logfile
```
smurawski: Enable-PSRemoting does a winrm quickconfig, so that's a redundancy.

lib/kitchen/driver/ec2.rb
tyler-ball: I have no idea - can I get some help from @smurawski or @salfunie?

smurawski: net user will add or modify existing accounts, so this would re-set the password to whatever test-kitchen thinks it should be for that user account.  This could break things if that user account was set up in application pools or something, but that just means you shouldn't run test-kitchen's transport as a user doing some other duty.

lib/kitchen/driver/ec2.rb
tyler-ball: What is the advantage in turning off the firewall?  In the simple case, all we need to be able to do is rdp/ssh into the instance from the workstation running TK, and the node needs to be able to communicate with the Chef Server so it can converge.

In advanced scenarios the target VM may need more connectivity but I say in that case the user needs to overwrite this default user data.

lib/kitchen/driver/ec2.rb
tyler-ball: I can't image the normal case needing to RDP from the target VM to something else, so I'll remove the client settings.

lib/kitchen/driver/ec2.rb
tyler-ball: Okay, I'm going to leave this behavior in as the default under the assumption that users won't have an image with the TK user pre-baked.  If they do, they can write their own custom user data.

lib/kitchen/driver/ec2.rb
smurawski: I'm not saying it is an advantage, just in one line we turn off the firewall and then we update a rule that doesn't matter because we turned the firewall off.

We should just need the below rule and not turn off the firewall with the first rule.

lib/kitchen/driver/ec2.rb
smurawski: yep.

smurawski:
LGTM :+1:

CHANGELOG.md
```diff
@@ -1,4 +1,19 @@
-## 0.9.4 / 2015-06-05
+## 0.10.0 / TBD
+
+### Bug Fixes
+
+### New Features
+
+* Added support for managing Windows hosts in AWS
```
zl4bv: @tyler-ball could you put our usernames against these?

tyler-ball: Absolutely!  Thanks for the reminder and all the great contributions.

zl4bv: Thanks for helping to pull this together!

afiune:
Just tested. This looks good.. :shipit: 

Thanks @tyler-ball for pushing this and make it happen (for real!) .. :smile:

PR #151, opened 2015-06-04
Query correct instance object for hostname fixes #148

### LINKED ISSUE #148: You broke Chef's Travis CI tests =)
@tyler-ball check out:

https://travis-ci.org/chef/chef/jobs/65344115

After the kitchen-ec2 release yesterday TK started trying to hit the private 172.x IPaddr of the virt rather than the public IP.

tyler-ball:
I introduced a regression where we use an old aws object to fetch connection information instead of the new object we queried to make sure it is ready

fixes https://github.com/test-kitchen/kitchen-ec2/issues/148

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -201,7 +201,6 @@ module Kitchen
         wait_until_ready(server, state)
 
         info("EC2 instance <#{state[:server_id]}> ready.")
-        state[:hostname] = hostname(server)
         instance.transport.connection(state).wait_until_ready
         create_ec2_json(state)
         debug("ec2:create '#{state[:hostname]}'")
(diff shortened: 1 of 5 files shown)


lib/kitchen/driver/ec2.rb
```diff
@@ -201,7 +201,6 @@ def create(state) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         wait_until_ready(server, state)
 
         info("EC2 instance <#{state[:server_id]}> ready.")
-        state[:hostname] = hostname(server)
```
tyler-ball: The problem is that I reassign the `server` object passed into `wait_until_ready`, so the server object here refers to the first iteration of it (before we continually updated it by querying for its hostname)

lib/kitchen/driver/ec2.rb
```diff
@@ -317,29 +316,40 @@ def tag_server(server)
...
-        rescue ::Aws::Waiters::Errors::WaiterFailed
-          error("Ran out of time waiting for the server with id [#{state[:server_id]}]" \
-            " to become ready, attempting to destroy it")
-          destroy(state)
-          raise
+        wait_with_destroy(server, state, "to become ready") do |aws_instance|
+          hostname = hostname(aws_instance, config[:interface])
+          # We aggressively store the hostname so if the process fails here
+          # we still have it, even if it will change later
+          state[:hostname] = hostname
```
tyler-ball: My fix is to store the hostname inside the method were we are updating the `server` (`aws_instance`) object

helgi:
:+1: pulling in this branch made things function again around the connections

PR #152, opened 2015-06-04
Pass updated server object back from wait_until_ready function

yoshiwaan:


tyler-ball:
Hey @yoshiwaan thanks for the PR, but this was originally filed as #148 and fixed in #151 that I just released as 0.9.5

PR #158, opened 2015-06-10
Remove useless log message

ustuehler:
AFAICS, the `Creating <#{state[:server_id]}>...` message will always output `Creating <>...` because `state[:server_id]` is not set yet. Otherwise, `return if state[:server_id]` would have returned early from the `create` method.


gmiranda23:
:+1:

tyler-ball:
You have no idea how many people asked about this.  Thanks for the PR!

Issue #159, opened 2015-06-17
Documentation - IAM policy document


zl4bv:
It would be nice to have documented the permissions that an IAM user will need to have in order to run kitchen-ec2.

Below is an IAM policy document that I have adapted from the one I [found](https://packer.io/docs/builders/amazon.html) on the Packer website. I've tested the policy, but someone else should probably confirm that this list makes sense for kitchen-ec2.

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ec2:AttachVolume",
                "ec2:CreateVolume",
                "ec2:DeleteVolume",
                "ec2:RunInstances",
                "ec2:TerminateInstances",
                "ec2:StopInstances",
                "ec2:DescribeVolumes",
                "ec2:DetachVolume",
                "ec2:DescribeInstances",
                "ec2:CreateTags"
            ],
            "Resource": "*"
        }
    ]
}
```

gmiranda23:
@zl4bv Anyone attempting to create ec2 instances would need permissions to manage the lifecycle of ec2 instances with the options they specify.  I'd hesitate to make a hard recommendation on the exact API calls they need to issue because it's not universal.  Example: a user might be testing a recipe that makes use of chef-provisioning-aws and calling out to services beyond ec2.  

We also expose a number of ec2 instance config options that make other calls beyond the basic ones listed (e.g. associating EIPs or options for Spot Instances) as PR's for them come in.  My concern is that any granular recommendation for exact API calls to allow would either go stale as the config option list grows or be too permissive/not permissive enough for different user scenarios.

Would it suffice to make reference to the fact that users should be authorized to manage any AWS resources they intend to modify or use and make a more general IAM recommendation like either using a managed policy such as arn:aws:iam::aws:policy/AmazonEC2FullAccess or a permissive policy like the one below and recommend users tailor it to their specific configuration requirements?

    {
      "Version": "2012-10-17",
      "Statement": [
        {
          "Action": "ec2:*",
          "Effect": "Allow",
          "Resource": "*"
        }
      ]
    }

zl4bv:
@gmiranda23 I was thinking of this as more of a base set of permissions for people to build upon than a concrete/complete list but I can see how easily the list could go stale and/or usage could be misconstrued by anyone trying to use it.

Having something more generic like your example would be sufficient - provided, as you say, there is a recommendation that users tailor it to their specific requirements.

[...]

PR #160, opened 2015-06-17
reference to required IAM settings

gmiranda23:
Addresses issue #159

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ instances. Use Amazon's cloud for your infrastructure testing!
 ## Requirements
 
 There are **no** external system requirements for this driver. However you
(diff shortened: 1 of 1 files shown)


zl4bv:
:+1:

tyler-ball:
I think this is a good start and I'll merge it.  But I'm going to leave https://github.com/test-kitchen/kitchen-ec2/issues/162 open because I think there are more permissions needed besides just the EC2 ones.  For example, we need to be able to query AMI information to check root_device status and look at the instance's console for the Windows case.  I would still like to have a more formal list of required actions (even if it is addition to `*` on ec2)

PR #161, opened 2015-06-20
Fix 'invalid char in json text' error

zl4bv:
After further testing and closer inspection of the Chef logs, it seems my change in #134 introduced a different kind of error:

```
[2015-06-18T21:34:00+00:00] DEBUG: No data to collect for plugin Virtualization. Continuing...
[2015-06-18T21:34:00+00:00] DEBUG: No hints present for azure.
[2015-06-18T21:35:20+00:00] ERROR: Could not parse hint file at C:\chef\ohai\hints/ec2.json: lexical error: invalid char in json text.
                                       ÿþ
                     (right here) ------^

[2015-06-18T21:35:20+00:00] DEBUG: has_ec2_mac? == false
[2015-06-18T21:35:20+00:00] DEBUG: looks_like_ec2? == false
```

Apparently the method I used inserts a [UTF-16 byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-16). 

The fix in this PR creates an empty file with no BOM.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -427,7 +427,7 @@ module Kitchen
 
       def create_ec2_json(state)
         if windows_os?
-          cmd = "mkdir \\etc\\chef\\ohai\\hints; echo $null >> \\etc\\chef\\ohai\\hints\\ec2.json"
+          cmd = "New-Item -Force C:\\chef\\ohai\\hints\\ec2.json -Type File"
         else
           cmd = "sudo mkdir -p /etc/chef/ohai/hints;sudo touch /etc/chef/ohai/hints/ec2.json"
         end


tyler-ball:
Assigning to myself so we will review this during bug triage

tyler-ball:
@zl4bv Nice catch and thanks for the PR.  I'm no Windows expert - will this work on Windows Server 2008?  I'm guessing it more depends on the version of Powershell present?

zl4bv:
As far as I know PowerShell has always used UTF-16 encoding (with BOM) as the default. [This](http://superuser.com/questions/327492/default-powershell-to-emitting-utf-8-instead-of-utf-16) thread leads me to believe this has been the case since at least PowerShell 2.0.

When you call the `New-Item` PowerShell Cmdlet without the `-Value` flag I guess it doesn't try to write to the file and therefore doesn't apply any encoding. With `$null >>` it writes (nothing) to the file but it still sets the encoding by that action.

zl4bv:
The `New-Item` Cmdlet was [introduced](https://technet.microsoft.com/en-us/library/hh849795(v=wps.620).aspx) in PowerShell 3.0 so it will work with Server 2008 and Server 2008 both as long as [Service Pack 1 installed](https://en.wikipedia.org/wiki/Windows_PowerShell#PowerShell_3.0) - which the images offered by Amazon do have installed.

tyler-ball:
I would like to get @smurawski's opinion, but I am :+1:

lib/kitchen/driver/ec2.rb
```diff
@@ -427,7 +427,7 @@ def hostname(server, interface_type = nil)
 
       def create_ec2_json(state)
         if windows_os?
-          cmd = "mkdir \\etc\\chef\\ohai\\hints; echo $null >> \\etc\\chef\\ohai\\hints\\ec2.json"
+          cmd = "New-Item -Force C:\\chef\\ohai\\hints\\ec2.json -Type File"
```
smurawski: The Type alias (for ItemType) wasn't introduced until PowerShell 3.  Server 2008R2 ships with PS 2 by default and this will work fine if you change the parameter to -ItemType (which will be valid from PowerShell 1 and up)

smurawski:
@zl4bv @tyler-ball I'd just make that one parameter name change and then I'm :+1:

zl4bv:
Thanks for the suggestion @smurawski :)

Did you still want me to change the parameter name @tyler-ball?

tyler-ball:
@zl4bv I did it when I merged it - thanks for the initial contribution!

Issue #147, opened 2015-06-04
Retrieve AMI IDs from the EC2 API


zl4bv:
This is a feature request intending the replace some or all of the functionality provided by the `amis.json` file. This is also a potential solution to #117.

Rather than creating a static mapping between a specific OS/image flavour, an EC2 region/AZ, and an AMI ID, I propose using the [describe-images](http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html#describe_images-instance_method) API endpoint to look up the AMI ID on-the-fly.

* You can search for AMIs by name using wildcards e.g. 'ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*' => latest Ubuntu 14.04 image
* AMI names often include the date which is why the wildcard is useful
* This ensures you get the latest version of an image (no more updating a list of mappings every month!)
* You can also filter by region, etc.

Either you could change `amis.json` to map a short name/alias to the filter parameters that get passed to `describe-images` e.g.

```json
{
  "ubuntu-14.04": [
    {"name": "ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"}
  ],
  "windows2012r2-base": [
    {"name": "Windows_Server-2012-R2_RTM-English-64Bit-Base-*"}
  ]
}
```

and/or you could leave it up to the user to provide the filter(s) in their `.kitchen.yml` file e.g.

```yaml
platforms:
  - name: windows2012r2
    driver:
      image:
        - name: Windows_Server-2012-R2_RTM-English-64Bit-Base-*
```

Thoughts @fnichol @tyler-ball ?

sqorbrody:
:+1:    I currently  am just manually changing the ami's from the following list, (https://cloud-images.ubuntu.com/trusty/current/)     This would be really nice

litjoco:
:+1: Agree this capability would be useful, and helpful for scenarios requiring private and frequently changing amis

proffalken:
+1 for this, it would make the lookup part of #199 redundant (although the "default to HVM+EBS would still stand)

freimer:
I'm not sure if this is related or not, but...
[...]

Issue #174, opened 2015-07-10
Default AMIs for Windows not available


emiliogarcia:
I can't find the default AMIs for Windows in us-east-1
https://github.com/test-kitchen/kitchen-ec2/blob/master/data/amis.json

I get:
No AMIs found matching your filter criteria
or not authorised to launch the AMI from the kitchen log.

Were those AMIs special? AFAIK WinRM is not enabled by default in the ones provided by Amazon. What does the Windows AMI needs to be able to work with Kitchen-ec2?

zl4bv:
Amazon release new Windows AMIs around once per month. The AMIs are then available for about six weeks after which you can no longer start new instances with those AMIs (the expectation being that you start new instances with this month's AMIs). Based on your error, it is most likely that the AMIs listed in amis.json are out of date now - which makes sense given they were last updated on June 5.

By default kitchen-ec2 [provides](https://github.com/test-kitchen/kitchen-ec2/blob/master/lib/kitchen/driver/ec2.rb#L476-490) user-data that enables WinRM for Windows guests. 

You'll need to make sure the security group of the test kitchen instance allows WinRM traffic on TCP port 5985. Also, either you need to provide a username and password for test-kitchen to connect to the instance with, or an SSH key and key ID so that kitchen-ec2 can decrypt the EC2 generated Windows password (with the default username Administrator).

tjnicholas:
It'd be easier to avoid stale AMI IDs if #177 was merged.

cheeseplus:
#177 is merged so calling this closed.

Issue #117, opened 2015-05-19
amis.json out of date


patcon:
Not sure when this happened, as I'm new, but noticed that the ubuntu-14.04 images were broken in all regions when searching on the official ami search. Confirmed that they aren't available here either: https://cloud-images.ubuntu.com/locator/ec2/

patcon:
Do these change often? We could write a quick script to scrape those reference sites for each platform

tyler-ball:
Yeah, they do change often.  I actually don't like having to support the `amis.json`.  I would rather write code to search for AMIs in real-time so we're not dependent on syncing these.

patcon:
Good point. If you don't feel like adding deps for scraping, using a scraperwiki sorta thing: http://morph.io/ (provides simple api with json responses for scrapers hosted on github). But then again, I'm partial to that service for opengov stuff :)

Also, your avatar is amazing.

tyler-ball:
I'll check out morph.io, and thanks!

whiteley:
This seems like a very similar need to https://github.com/chef/chef-provisioning-aws/pull/197 which I suggested a fix using the `ubuntu_ami` gem. If that seems like a reasonable dependency, I'd be happy to make a PR for it.

tyler-ball:
I like your other PR, so please make one for kitchen-ec2!  Thanks!

gsaslis:
yep - definitely out of date. 

ubuntu-14.04 [ami id](https://github.com/test-kitchen/kitchen-ec2/blob/258b658911bf6999dd020844fb38f2ce28b3a6f4/data/amis.json#L39) in eu-west-1 is also out of date

cheeseplus:
We've adding looking up and updated AMIs - please open new issues with any per-plaftform updates independently.

PR #177, opened 2015-07-16
Add support for looking up AMIs with the EC2 API

zl4bv:
Addresses #147. When a hash of filters are provided for the image_id it will try to find the AMI with the most recent creation date that matches the supplied filters.

Example usage (.kitchen.yml):
```yaml
platforms:
  - name: windows2012r2
    driver:
      image_id:
        name: Windows_Server-2012-R2_RTM-English-64Bit-Base-*
```

Example usage (amis.json):
```json
{
  "regions": {
    "ap-southeast-2": {
      "windows-2012r2": {
        "name": "Windows_Server-2012-R2_RTM-English-64Bit-Base-*"
      }
    }
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ token) through config.  See the `aws_access_key_id` and `aws_secret_access_key`
 config sections below to see how to specify these in your .kitchen.yml or
 through environment variables.  If you would like to specify your session token
 use the environment variable `AWS_SESSION_TOKEN`.
-1. The shared credentials ini file at `~/.aws/credentials`.  You can specify 
+1. The shared credentials ini file at `~/.aws/credentials`.  You can specify
 multiple profiles in this file and select one with the `AWS_PROFILE`
 environment variable or the `shared_credentials_profile` driver config.  Read
 [this][credentials_docs] for more information.
@@ -89,7 +89,7 @@ environment variable or the `shared_credentials_profile` driver config.  Read
 metadata service to discover the local instance's IAM instance profile.
 
 This precedence order is taken from http://docs.aws.amazon.com/sdkforruby/api/index.html#Configuration
-  
+
 The first method attempted that works will be used.  IE, if you want to auth
 using the instance profile, you must not set any of the access key configs
 or environment variables, and you must not specify a `~/.aws/credentials`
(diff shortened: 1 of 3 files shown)


willejs:
Is there any update on this? The outdated ami id's keep breaking our CI...

willejs:
Is there an update on this? Would love to see it merged!

puckel:
Hi,

:+1: for this PR, it'll be very useful. Maybe add an environment variable for image_id too.

Thanks

zl4bv:
Hey @puckel, if you want to set the image_id with an environment variable then you can add something like the following to your `.kitchen.yml` file:

```yaml
platforms:
  - name: blah
    driver:
      image_id: "<%= ENV['IMAGE_ID'] %>"
```

where the environment variable `IMAGE_ID` has your desired AMI ID. This will work without the pull request.

willejs:
@zl4bv I think the idea here is that we want to use the latest and greatest AMI ID for a given os & version . Maintaining current AMI IDs is a nightmare as they do not last forever.

tjnicholas:
@tyler-ball any estimate when this might get merged and/or released? It looks like it'll address several open issues. #174 #117 #147

tyler-ball:
Sorry this got dropped.  The code and tests look good, but I see a few missing features to add before this can be merged.

First, while it is nice to use an existing key I think that will be confusing in the long run.  Instead of overloading `image_id` to take either a string or a hash I think we should just introduce a new config key.  Call it `image_search` and have it take in a hash.  If both `image_id` and `image_search` are provided then give `image_id` precedence.  Because `image_id` is currently defaulted you will need to move the defaulting into the [`finalize_config`](https://github.com/test-kitchen/kitchen-ec2/blob/master/lib/kitchen/driver/ec2.rb#L162) method.  It will look something like

```ruby
if config[:image_id].nil? && config[:image_search].nil?
  config[:image_id] = instance.driver.default_ami
end
```

Second, this PR needs to also update the [README](https://github.com/test-kitchen/kitchen-ec2#image_id) to outline how to leverage this new logic.

We should leave the existing `amis.json` logic alone but I think that is another issue we could address in the 1.0 update.  Ideally, I would like to get rid of it and use a default image of the latest ubuntu image (using the search tool here).  But that should be another PR.  My time has been prioritized off of Test Kitchen so if anyone wants to submit that PR I would be happy to review and merge it.

willejs:
@zl4bv do you think your going to get round to fixing this with tylers comments?

zl4bv:
@willejs sorry, yes, I will try to get it done some time this week.

zl4bv:
Hey @tyler-ball (and other collaborators), I've made the suggested changes. Do you think it is ready to be merged now?

willejs:
YES! Can you cut a release please @tyler-ball

puckel:
Hi,
When the next release is planned?
Thanks

willejs:
new release FTW!

Issue #172, opened 2015-07-06
Support empheral storage


atward:
I currently can't see a way to attach ephemeral drives. 

Using the following config, which is valid for cloudformation and block_device_mappings in the aws ruby sdk:

```
driver:
  name: ec2
  instance_type: m3.large
  block_device_mappings:
    - ebs_device_name: /dev/sdb
      ebs_virtual_name: ephemeral0
```

Produces error:

```
$ kitchen list
/Users/adamward/.chefdk/gem/ruby/2.1.0/gems/kitchen-ec2-0.9.5/lib/kitchen/driver/ec2.rb:117:in `block (2 levels) in <class:Ec2>': Every :block_device_mapping must include the keys :ebs_volume_size, :ebs_delete_on_termination and :ebs_device_name (RuntimeError)
```

Relates to #164 - as being able to provide the direct hash should solve this issue.

cheeseplus:
This has been in for a while now https://github.com/test-kitchen/kitchen-ec2/#disk-configuration, closing

PR #190, opened 2015-08-18
Will provide support for ephemeral block_device_mappings

erkolson:
Resolves issue #172 

Example of ephemeral drives used on the same instance in conjunction with an ebs device, in kitchen.yml:
```yaml
      block_device_mappings:
        - ebs_device_name: /dev/xvda
          ebs_volume_type: gp2
          ebs_volume_size: 8
          ebs_delete_on_termination: true
        - device_name: /dev/xvdb
          virtual_name: ephemeral0
        - device_name: /dev/xvdc
          virtual_name: ephemeral1
```

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -116,12 +116,16 @@ module Kitchen
       validations[:block_device_mappings] = lambda do |_attr, val, _driver|
         unless val.nil?
           val.each do |bdm|
-            unless bdm.keys.include?(:ebs_volume_size) &&
+            if bdm.keys.include?(:ebs_volume_size) &&
                 bdm.keys.include?(:ebs_delete_on_termination) &&
                 bdm.keys.include?(:ebs_device_name)
+            elsif bdm.keys.include?(:device_name) &&
+                  bdm.keys.include?(:virtual_name)
+            else
               raise "Every :block_device_mapping must include the keys :ebs_volume_size, " \
-                ":ebs_delete_on_termination and :ebs_device_name"
-            end
+                ":ebs_delete_on_termination and :ebs_device_name for EBS devices \
+                OR device_name, virtual_name for Ephemeral devices"
+            endd
           end
         end
       end
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -120,7 +120,7 @@ module Kitchen
                 bdm.keys.include?(:ebs_delete_on_termination) &&
                 bdm.keys.include?(:ebs_device_name)
             elsif bdm.keys.include?(:device_name) &&
-                  bdm.keys.include?(:virtual_name)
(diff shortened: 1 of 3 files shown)


kingpong:
+1

grosendorf:
:+1:

dougalb:
+1

dzabel:
Would be great to see this in next release

cheeseplus:
Could we get this rebased so the PR gets tested using the latest travis configs?

erkolson:
@cheeseplus, I think that covers it.

lib/kitchen/driver/ec2.rb
```diff
@@ -116,11 +116,15 @@ def self.validation_warn(driver, old_key, new_key)
       validations[:block_device_mappings] = lambda do |_attr, val, _driver|
         unless val.nil?
           val.each do |bdm|
-            unless bdm.keys.include?(:ebs_volume_size) &&
+            if bdm.keys.include?(:ebs_volume_size) &&
```
jkeiser: Why change  unless to if here? The "empty if" pattern is a little hard to read.

jkeiser:
@erkolson I have submitted #223 which fixes your problem by allowing arbitrary values to be passed to `block_device_mappings`. To do what you suggested in your example above, you can do this:

```yaml
      block_device_mappings:
        - device_name: /dev/xvda
          ebs:
            volume_type: gp2
            volume_size: 8
            delete_on_termination: true
        - device_name: /dev/xvdb
          virtual_name: ephemeral0
        - device_name: /dev/xvdc
          virtual_name: ephemeral1
```

jkeiser:
If you get some time to try that out, I'd super appreciate it! If it works for you, we can close this one.

jkeiser:
I believe this is fixed in master now (with the aforementioned patch having been merged).

erkolson:
I don't see it in master... how should I test?

jkeiser:
It was this one: https://github.com/test-kitchen/kitchen-ec2/commit/5e92115b69b26beaaf33a22be56d43ded7dfde81

You can either `rake install` from latest master, or you can `gem install kitchen-ec2-1.0.0.beta.1`.

erkolson:
Thanks, I'll test on Monday.  I've been working on a production issue all night and don't feel like doing much more today :-/

TheLinuxNinja:
I just came across this while trying to find why some of our .kitchen.yml files don't work anymore. Seems that we were able to have ephemerals at one time, then they were broken. I see the last activity here is over a month ago. Where does it stand?

erkolson:
@jkeiser, Sorry for the delay.  I was able to test 1.0.0.beta.1 yesterday.  It works for me.  Is this already in the 1.0.0 release?

jkeiser:
1.0.0 should have this, yep!

roshi619:
@jkeiser .. it seems like this is not yet available within 1.0.0. Could you check and merge accordingly.

PR #197, opened 2015-09-29
Add support for looking up Private DNS Name for hostname

mekf:
In some circumstances - for example proxy settings that are configurable on hostnames - the `private_dns_name` is preferred over the `private_ip_address`

When `private_dns` is provided as the option in `.kitchen.yml`, the value of `private_dns_name` will be returned

```ruby
driver:
  interface: private_dns
```

When this option is unset, the `private_ip_address` will be returned as in the current implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,13 +325,21 @@ If you don't set this it will default to whatever DHCP address EC2 hands out.
 
 ### interface
 
-The place from which to derive the hostname for communicating with the instance.  May be `dns`, `public` or `private`.  If this is unset, the driver will derive the hostname by failing back in the following order:
+The place from which to derive the hostname for communicating with the instance.  May be `dns`, `public`, `private`, `private_dns`.  If this is unset, the driver will derive the hostname by failing back in the following order:
 
 1. DNS Name
 2. Public IP Address
 3. Private IP Address
+4. Private DNS Name
 
-The default is unset.
+The default is unset. Under normal circumstances, the lookup will return the `Private IP Address`.
+
+If the `Private DNS Name` is preferred over the private IP, it must be specified in the `.kitchen.yml` file
+
+```ruby
+driver:
+  interface: private_dns
+```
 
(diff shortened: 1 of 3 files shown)


mekf:
\cc @zl4bv @tjnicholas

jkeiser:
@mekf is the private dns name ever preferred over the private ip? Do we want to inject this third in the priority instead of fourth?

zl4bv:
> is the private dns name ever preferred over the private ip?

Given a user is behind a proxy
And the user's no_proxy list is set to `.compute.internal` (because [IP prefixes are not supported](https://github.com/nahi/httpclient/blob/f60f6bad0a516b08052c5322eef21ffd9ff7ba08/lib/httpclient.rb#L527))
And the kitchen-ec2 instance is also behind the proxy
And the transport is set to WinRM

When kitchen-ec2 connects to the instance

Then it should connect to the instance using the private DNS address
And the WinRM requests should not go through the proxy

jkeiser:
I'm unclear whether that is a more common or less common case than the opposite. Well, :+1: at any rate, seems reasonable.

zl4bv:
The flipside is that the `compute.internal.` zone is generally only resolvable inside AWS. So a host outside of AWS would not be able connect to a test instance using private DNS even if the correct connectivity/routing was in place.

Certainly not sure which case is more common, but the "play it safe" option in my opinion would be to keep the current order with private IP address preferred over private DNS.

jkeiser:
Fair enough! Paging @test-kitchen/maintainers @test-kitchen/kitchen-ec2-maintainers

proffalken:
Code looks fine to me, approve of the ordering as well - I've had some real issues with Ansible when trying to SSH via private_dns so defaulting to the IP is the sensible thing to do IMHO.

+1

mekf:
thanks @jkeiser @zl4bv

PR #208, opened 2015-11-10
Allow configuring retry_limit in Aws.config

jlyheden:
Setting this value high in .kitchen.yml allows us to run tons of suites in parallell without failing on api throttling


jkeiser:
:+1:

cheeseplus:
+1

PR #221, opened 2016-02-10
Dynamically find default images for many platforms

jkeiser:
- Use search to find latest official AMI of the given OS and version on your region:
  - ubuntu[-14[.04]]
  - windows[-2008|-2012|-2012r2[sp1|sp2|rtm]]
  - rhel[-major[.minor]]
  - debian[-major]
  - centos[-major[.minor]]
  - freebsd[-major[.minor]]
  - fedora[-major]
- Prefer images by (in order of priority):
  1. Latest version
  2. SSD support over non-SSD
  3. HVM (modern standard) over paravirtual
  4. 64-bit over 32-bit
  5. More recent image creation date
- Default username: Auto-detect ssh username from image name
- Default instance_type: t2.micro or t1.micro depending on image type (t2.micro for hvm, t1.micro for paravirtual)
- Do not default availability zone (default of "a" doesn't work in all regions, and it's not necessary to pass AZ to create an instance)


README.md
```diff
@@ -9,84 +9,147 @@ A [Test Kitchen][kitchenci] Driver for Amazon EC2.
 This driver uses the [aws sdk gem][aws_sdk_gem] to provision and destroy EC2
 instances. Use Amazon's cloud for your infrastructure testing!
 
+## Initial Setup
+
+To get started, you need to install the software and set up your credentials and SSH key. Some of these steps you have probably already done, but we include them here for completeness.
+
+1. Install the latest ChefDK and put it in your path.
```
adamleff: The TK maintainers are trying to ensure the project is agnostic to the use case. So, we should probably change this line to add "or ensure the test-kitchen gem is installed on your workstation"?

README.md
```diff
@@ -9,84 +9,147 @@ A [Test Kitchen][kitchenci] Driver for Amazon EC2.
...
+2. From this repository, type `bundle install; bundle exec rake install` to install the latest version of the driver.
```
adamleff: We shouldn't require (or at least shouldn't encourage) users to clone the repo just to be able to use this gem. Can we change this to using `gem install` (or `chef gem install` if using Chef DK)?

jkeiser: The instructions in this readme cover the code in `master`--the things it says are not true of earlier versions. I think it'd be best to hold off and change it to `gem install` after we release the code. Once we check this in, we can release an alpha or something, and point at that ...

lib/kitchen/driver/aws/standard_platform.rb
```diff
@@ -0,0 +1,216 @@
...
+        #
+        # @return [String] The image ID (e.g. ami-213984723)
+        def find_image(image_search)
+          driver.debug("Searching for images matching #{image_search} ...")
+          # Convert to ec2 search format (pairs of name+values)
+          filters = image_search.map do |key, value|
+            { name: key.to_s, values: Array(value).map { |v| v.to_s } }
+          end
+
+          # We prefer most recent first
```
adamleff: This block from 83-110 is so cool. Neat implementation.

README.md
jkeiser: Done!

lib/kitchen/driver/ec2.rb
```diff
@@ -57,20 +66,20 @@ class Ec2 < Kitchen::Driver::Base # rubocop:disable Metrics/ClassLength
       default_config :price,              nil
       default_config :retryable_tries,    60
       default_config :retryable_sleep,    5
-      default_config :aws_access_key_id,  nil
+      default_config :aws_access_key_id,  ENV["AWS_SSH_KEY_ID"]
```
adamleff: I think I'm misunderstanding this change (and the one below in 72). The SSH key ID is not the same as the access_key. Can you help me understand this one?

lib/kitchen/driver/ec2.rb
```diff
@@ -251,36 +235,60 @@ def destroy(state)
...
+      def desired_platform
+        @desired_platform ||= begin
+          platform = Aws::StandardPlatform.from_platform_string(self, instance.platform.name)
+          if platform
+            debug("platform name #{instance.platform.name} appears to be a standard platform." \
+                  " Searching for #{platform} ...")
+          else
+            debug("platform name #{instance.platform.name} does not have a valid os." \
+                  " Searching for latest stable ubuntu ...")
+            platform = Aws::StandardPlatform.from_platform_string(self, "ubuntu")
```
adamleff: I might be in the minority, but it feels weird to default to ubuntu if we can't figure out what the user wants. Shouldn't we raise an exception here, or at least be very, very vocal about us defaulting to a particular OS?

lib/kitchen/driver/ec2.rb
```diff
@@ -311,13 +319,16 @@ def copy_deprecated_configs(state)
...
-          # This duplicated old behavior but I hate amis.json
-          ami_username = amis["usernames"][instance.platform.name]
-          state[:username] = ami_username if ami_username
+        state[:username] = config[:username]
+        unless instance.transport[:username] == instance.transport.class.defaults[:username]
+          state[:username] ||= instance.transport[:username]
+        end
+        if !state[:username]
+          if actual_platform
+            debug("No SSH username specified: using default username #{actual_platform.username} " \
```
adamleff: Given how often people get hung up on authentication issues in TK, I'd actually `info` this. :)

lib/kitchen/driver/ec2.rb
proffalken: Agreed - Ubuntu does seem to be the darling of the Linux World, however, I would suggest that we default to AWS linux  given that this is also Amazon's default.

This would mean that there are no surprises when launching a VM if you don't specify the distro as you get the same distro that you would get if you just kept clicking "next" via the AWS UI

lib/kitchen/driver/ec2.rb
proffalken: +1

lib/kitchen/driver/ec2.rb
jkeiser: It's wrong :)

lib/kitchen/driver/ec2.rb
jkeiser: This is what we already do though, isn't it? If it isn't, I will totally get rid of it.

lib/kitchen/driver/ec2.rb
jkeiser: I feel like the right thing to do here is to have a better error message when SSH retries. It is total crap that we don't say why it's failing and what username we are using.

lib/kitchen/driver/ec2.rb
adamleff: I don't believe so. I can't find where we default to ubuntu currently.

lib/kitchen/driver/ec2.rb
adamleff: I totally agree, but there's no harm in "info"ing this too to give even more clues to the user.

jkeiser: Done! Not this specifically, but all auto-detection is printed out at one point or another.

adamleff:
LGTM. @jkeiser this is some sooper-sweet stuff. Thanks for doing it.

Paging @test-kitchen/kitchen-ec2-maintainers

README.md
```diff
@@ -9,84 +9,147 @@ A [Test Kitchen][kitchenci] Driver for Amazon EC2.
...
+platforms:
+  # The latest stable minor+patch release of rhel 6
+  - name: rhel-6
+  # The latest patch release of CentOS 6.3
+  - name: centos-6.3
+  # 32-bit version of latest major+minor+patch release of Ubuntu
+  - name: ubuntu-i386
+  # 32-bit version of Debian 6
+  - name: debian-6-i386
+  # Latest 64-bit stable minor release of freebsd 10
```
thommay: do you mean 64bit? cos the name says `i386`

thommay:
One teensy nit otherwise :+1:

lamont-granquist:
Would changing the priority of "64-bit vs 32-bit" to second change anything?  HVM is also most likely more important than SSDs, but I suspect most of the SSD images are HVM 64-bit anyway.

And have you tried to converge chef-client on a t1.micro recently with a non-trivial runlist?  The last time I tried running chef-client on a t1.micro it got a few minutes into a first converge and then it triggered the CPU governor which took away 99% of the CPU and it took the first converge a few more hours to run when it only had 1% of a CPU left.  On a null run-list they probably finish fast enough that the governor does not kick on.  That was several years ago so things might have changed, but they were not fit for purpose last time I checked.

jkeiser:
@lamont-granquist hmm, I think doing amd64 > i386 earlier makes a lot of sense from a consistency point of view.

For the priority of HVM over SSD, do you think the SSD generally matters more, or the instance_type? That's the main thing you are selecting--the ability to actually run on most instance types.

As far as `t1.micro`, what do you suggest is a good default? I was super reluctant to bring it out of the free tier, but paravirtual images literally do not work on `t1.micro`.

lamont-granquist:
What happens if you remove paravirt/t1.micro images completely?  How much coverage do you lose?

jkeiser:
You lose some OS's and versions (can't recall which). Certainly nothing recent. But I can't see a reason to just not support them at all, if we have a way ...

lamont-granquist:
Yeah, but last time I tried the t1.micros they were pretty much unfit for purpose.  That's kinda the point I'm not sure you're 'supporting' them just by adding the config...

PR #222, opened 2016-02-12
Ephemeral drives

jkeiser:
Rebase of #190


jkeiser:
Actually, this is obsoleted by the fact that block_device_mappings is now verbatim, hooray!

PR #227, opened 2016-02-22
Add support for "Assume Role" credentials

yyuu:
By security reason, basically I don't like to store AWS credentials with strong privileges as much as possible. Luckily there's STS's `AssumeRole` API in AWS and it allows users to generate temporary keypair which will be expired automatically in 1 hour.

http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html

This patch should use temporary credentials generated by `AssumeRole` if `assume_role_arn` and `assume_role_session_name` are given in the configurations.


iiro:
+1

ehaselwanter:
+1

yyuu:
Any updates? Or, should I implement some more tests or something?

JaBurd:
@yyuu we use a "renew token" script outside our converges to generate the temp token. The temp profile is used for spinning up the box initially. Not integrated, nor clean and easy but enables the functionality needed.

Maniacal:
I work for a company in which you must 'assume role' to get anything done in AWS.  This blocks us from using kitchen-ec2.  This feature would be welcomed.

iiro:
Like @Maniacal I also work for a company where you must use the AssumeRole with AWS. This would be crucial for us also.

cheeseplus:
If we can get this rebased I don't have any issues with adding the functionality but will need to review again.

yyuu:
I've just rebased on to the latest master branch. Please give it a look.

gdavison:
@Maniacal @iiro A colleague of mine wrote: https://github.com/vancluever/aws-runas. This might help with needing to assume a role

cheeseplus:
We merged this and then reverted due to some issues we were having (which may have been transient due to other upstream issues). Github is cranky about us reverting the revert so I'm going to create another PR with the patch set, test, and hopefully get that merged soon.

PR #229, opened 2016-02-24
Rename price -> spot_price, fix rubocop

jkeiser:


mwrock:
:+1:

lib/kitchen/driver/ec2.rb
```diff
@@ -62,7 +62,7 @@ class Ec2 < Kitchen::Driver::Base # rubocop:disable Metrics/ClassLength
       end
       default_config :private_ip_address, nil
       default_config :iam_profile_name,   nil
-      default_config :price,              nil
+      default_config :spot_price,              nil
```
tas50: nitpick here, but it's not aligned anymore

tas50:
spotprice is the right thing to do, but we need to bump this a major for this one since it's going to break anyone using price.

jkeiser:
@tas50 `price` is an entirely unreleased change :) But we're bumping major anyway.

proffalken:
+1 here (for the change and the version bump...)

PR #235, opened 2016-03-07
Adding support for tenancy parameter in placement config.

jcastillocano:
Only "default" and "dedicated" values are accepted. If you think is worthy to have also "host" value (http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/dedicated-instance.html#dedicated-howitworks) but being honest I've never seen that value in an instance config. This parameter works in pair with availability zone to define the placement attribute.


jcastillocano:
Hi @tyler-ball 

do you need anything else from my side to improve this PR in order to be merged?

Kind regards

jcastillocano:
@cheeseplus @jkeiser 

Could you pleaes help me to merge this PR? 

Regards

corecanarias:
Hi, can we have this merged?

cheeseplus:
👍

lib/kitchen/driver/ec2.rb
```diff
@@ -79,6 +79,7 @@ class Ec2 < Kitchen::Driver::Base # rubocop:disable Metrics/ClassLength
       default_config :interface,           nil
       default_config :http_proxy,          ENV["HTTPS_PROXY"] || ENV["HTTP_PROXY"]
       default_config :retry_limit,         3
+      default_config :tenancy,             "dedicated"
```
corecanarias: Not sure if this is the real default value when tenancy is not specified. I guess that by default should be "default". Guess not everybody wants to use dedicated instances.

jcastillocano:
Hi, I just merged from master. Could we have this merged?

PR #240, opened 2016-03-25
Add synced_folders feature

aogail:
I am interested in adding a "synced folders" feature to kitchen-ec2. This is meant to be similar to the kitchen-vagrant synced_folders feature: you specify a local path and a remote path, and kitchen will copy the local path to the remote path on the VM.

So far I've just written an implementation that works for me. Before I spend time on adding tests and whatnot, I wanted to open this PR to check whether this feature would be a welcome addition to kitchen-ec2. :)

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -79,6 +79,7 @@ module Kitchen
       default_config :interface,           nil
       default_config :http_proxy,          ENV["HTTPS_PROXY"] || ENV["HTTP_PROXY"]
       default_config :retry_limit,         3
+      default_config :synced_folders, []
 
       required_config :aws_ssh_key_id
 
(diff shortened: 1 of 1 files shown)


mwrock:
I think this is a good idea. I'd actually think that the implementation here would ideally be in the driver base class so all drivers could potentially benefit. No reason why they couldn't since there is nothing fancy here and you are simply leveraging the transport.

Vagrant is the only driver that natively supports this and I'd think we'd just want to have it override the base behavior with its own.

smurawski:
I agree with @mwrock - this type of capability should be in the base driver class.

aogail:
@mwrock @smurawski Putting the functionality in the base driver class sounds good to me. In practical terms, were you thinking of having a "sync folders" method in `Kitchen::Driver::Base` that drivers would call in their `create` methods? Or something else?

mwrock:
that sounds reasonable to me.

aogail:
@mwrock @smurawski I opened test-kitchen/test-kitchen#979 to implement most of this in the base driver, per this discussion.

mwrock:
very cool. I'll close here.

PR #251, opened 2016-05-11
fixes #250 and provides the option to set ssl_peer_verify to false

### LINKED ISSUE #250: driver fails on ssl-verify when run on a ssl intercepting network
More and more commonly, users are wanting to leverage these tools inside of an enterprise network that has SSL Intercepting features enabled (see https://bto.bluecoat.com/webguides/proxysg/security_first_steps/Content/Solutions/SSL/ssl_solution.htm for one such example).

On a network with this enabled, attempts to verify the SSL certificates will fail, as the cert will not show as valid (as the cert is from the proxy device and not the actual destination).

Example output of running kitchen inside such a network:

```
-----> Starting Kitchen (v1.7.3)
D      winrm requested, loading winrm gem (["~> 1.6"])
D      winrm is loaded.
D      winrm-fs requested, loading winrm-fs gem (["~> 0.4.1"])
D      winrm-fs is loaded.
-----> Creating <default-windows-2012r2>...
D      ------Exception-------
D      Class: Kitchen::ActionFailed
D      Message: Failed to complete #create action: [SSL_connect returned=1 errno=0 state=error: certificate verify failed]
D      ---Nested Exception---
D      Class: Seahorse::Client::NetworkingError
D      Message: SSL_connect returned=1 errno=0 state=error: certificate verify failed
D      ------Backtrace-------
```

Ideally, there should be some type of setting in the driver to bypass ssl verification for such cases.

mwrock:


coderanger:
👎 These options are a scourge and too much of a footgun IMO.

adamleff:
:+1: but could we change the parameter name so it's extremely clear what it's for? It feels too generic now which may confuse users. 

How about `verify_aws_ssl_peer`, `aws_api_ssl_verify`, or similar?

mwrock:
@adamleff in a slack conversation we have been discussing the merits of other routs tro achieve the same end but without flipping the power button on security.

The thinking is that if we can provide guidance to users on how to manage their cacert.pem store with the fake root cert used in an environment as described in #250, thats a more "legit" approach. I want to keep this open while we explore that. Question is can users actually make it work in possibly sophisticated proxied environments where they may need to feed a chain of certs to the openssl cert file.

A carte-blanche verify_none may be the lesser evil in some dev/test environments.

coderanger:
If we do add something like this as an escape hatch, I would make the name something that makes it clear this is security risk. "disable TLS verification" sounds bad, but like "click through a Windows warning I've been desensitized to" bad :-/ Maybe just have a boolean for `disable_tls_security`?

adamleff:
Yeah, I'm not crazy about having an option like this either, but we do have users where this may be the right solution. 

We could also look at doing something like `kitchen ssl fetch` and `kitchen ssl verify` (a la knife) that each driver can implement depending on needs. 

Proxies gonna prox. A "quick fix" here is to just make this change but also print out a HUGE warning each time kitchen is run explaining what's happening and encouraging them to visit a URL to a doc that explains how to fix this while still maintaining TLS verification.

cheeseplus:
If we can get this rebased and an appropriate warning and option name selected this still seems like a good thing.

cheeseplus:
Can we get a rebase? are we still 👍  on this?

mwrock:
rebased. I'm still +1.

cheeseplus:
@mwrock merged some other PRs and now we need a rebase here - will get this as soon as you land one

cheeseplus:
The oddest thing is that now it says the merge has no conflicts which may be due to my merge/revert so I guess we don't need a rebase after all.

PR #252, opened 2016-05-13
Add notes for AMI image name requirements

freimer:
Add notes for AMI naming convention requirements for private custom AMIs.  May be related to issue #147


jrnt30:
Just ran into something similar from your comment on #147 as well.  There is a slight caveat, the "windows" can be anywhere in the name I found and case does not matter.  

Regardless, I think outlining this would be a big help and would have saved me several hours

cheeseplus:
Thanks for helping clear this up, +1

PR #257, opened 2016-06-06
Make tags optional for clients without IAM rights to CreateTags

freimer:
Some EC2 users do not have the ability to CreateTags, because there are no granular controls around that (yet).  For instance, a user with CreateTags can wipe or change any tag on any resource in an account, which is not good.  This simply does not call the CreateTags API if there are no tags defined.  By default, there is a created-by tag, but if tag: is set in the config with no keys it will be nil, and the if block will not get executed.  Please merge this if it looks OK.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -355,11 +355,13 @@ module Kitchen
       end
 
       def tag_server(server)
-        tags = []
-        config[:tags].each do |k, v|
-          tags << { :key => k, :value => v }
+        if config[:tags]
+          tags = []
+          config[:tags].each do |k, v|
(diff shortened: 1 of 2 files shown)


smurawski:
Thanks for the PR @freimer.  The fix looks good to me, but it'd be great to get a spec test as well, to make sure we don't revert the desired behavior in a future change.

freimer:
I would love to, but I don't know how to do that yet.  I'm still relatively new to all of this ;-)

smurawski:
👍

freimer:
Any news on this?

cheeseplus:
+1

stevenoneill:
+1

stevenoneill:
woohoo!

PR #260, opened 2016-06-22
propagate tags to volumes

mrbobbytables:
This PR propagates tags to volumes.

Similar to the instance being created before it's usable, volumes may not be available to the instance immediately. 

Through some trial and error I found `ec2.client.describe_volumes` to consistently return what volumes should be attached to the instance before enumerating them through the instance itself (`server.volumes`). 

`wait_until_volumes_ready` mimics `wait_until_ready` while comparing what the client can enumerate about the instance's volumes to the instance itself. Once they agree -- proceed and tag the volumes.


gsfjohnson:
Volume tagging would be nice to have.  Any updates on when/if this will be merged?

cheeseplus:
LGTM and if we can get another reviewer then we'll get this in for the next release.

PR #264, opened 2016-07-19
Fix autoload race in Aws::EC2::*

jkeiser:
by accessing the resources we need in driver initialize.

Tests for this are hard to write because they have to run separately. Here is the test I ran:

```
require "kitchen/driver/ec2"

ec2 = Kitchen::Driver::Ec2.new({})
threads = (1..1000).map do
  Thread.new do
    begin
      ec2.ec2.resource
    rescue
      STDERR.puts "Noooo #{$!}"
    end
  end
end
threads.each { |t| t.join }
```

If I run this *before* my patch, it prints out "Nooooo uninitialized constant Aws::EC2::Resource" multiple times. After the patch, it doesn't print it out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ gemspec
 gem "test-kitchen"
 gem "winrm-transport"
 gem "winrm-fs"
+gem "activesupport", "~> 4.0"
 
 group :test do
   gem "rake"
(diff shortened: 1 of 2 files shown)


ksubrama:
:+1:

mwrock:
:+1:

PR #270, opened 2016-08-08
Fix AWS Ruby SDK autoload for all time

jkeiser:
Fix #264 was not complete enough: other things still get autoloaded and have caused multiple failures in our CI suites. Tested by running the repro in #264.

This seems not to be an issue in Ruby 2.3.1, from my tests, so I've capped the eager_autoload workaround at 2.2 and below.


mwrock:
:+1:

cheeseplus:
+1

Issue #225, opened 2016-02-15
transport.username is not respected


alexpop:
With code in master(c97798142d639ff3cff435b5516eb3be00bce6a3), I'm using this kitchen.ec2.yml config to converge `us-west-1` instances:

```
platforms:
  - name: centos-6
    driver:
      image_id: ami-e73ccda3
    transport:
      username: root
    provisioner:
      sudo: false
```

Kitchen is attempting to login with the default username(`centos`) instead of the configured one(`root`):

```
$ bundle exec kitchen converge
-----> Starting Kitchen (v1.5.0)
-----> Creating <default-centos-6>...
       Detected platform: centos version 5 on x86_64. Instance Type: m3.medium. Default username: centos (default).
...SNIP...
       Waited 25/300s for instance <i-3c16fb89> to become ready.
       EC2 instance <i-3c16fb89> ready.
       Waiting for SSH service on ec2-54-183-129-223.us-west-1.compute.amazonaws.com:22, retrying in 3 seconds
       Waiting for SSH service on ec2-54-183-129-223.us-west-1.compute.amazonaws.com:22, retrying in 3 seconds
       Waiting for SSH service on ec2-54-183-129-223.us-west-1.compute.amazonaws.com:22, retrying in 3 seconds
...
```

After I cancel the `Waiting for SSH` loop, the contents of `.kitchen/default-centos-6.yml` is:

```

---
username: centos
server_id: i-3c16fb89
hostname: ec2-54-183-129-223.us-west-1.compute.amazonaws.com
```

I'm able to converge fine with the same config if I use the kitchen-ec2 master ref 0e363e068c8d03978358b31b19b1f9dc09fac25d

jkeiser:
Can you add `-l debug` so we can find out what it's actually using to connect? The default username isn't necessarily what it will use to connect. We really need to get test-kitchen to give more information about how it's connecting.

jkeiser:
In my tests, furthermore, it is in fact using `transport.username` to connect:

```
driver:
  name: ec2
  aws_ssh_key_id: jkeiser
  region: us-west-2

transport:
  ssh_key: ~/.ssh/jkeiser

# This is a big list of automatically supported platforms
platforms:
  - name: centos-6.5
    transport:
      username: blah
```

(Well, in this case it's failing to connect, but when I run with `debug` on it's clearly using `blah`.)

proffalken:
+1 for further information including debug logs, I'm struggling to see how
[...]

PR #273, opened 2016-08-30
Don't break if we're using a custom "platform" AMI

hynd:
If i'm using an private image_id whose AMI Name doesn't match one of the "standard_platform"s (ie; "oel-7.2"), i can't seem to override the username.  Instead, i get the following error:

```
>>>>>> ------Exception-------
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #create action: [undefined method `username' for nil:NilClass] on default-oel-72
>>>>>> ----------------------
```

which - i think - is because [actual_platform](https://github.com/test-kitchen/kitchen-ec2/blob/119001d1ad2e7879151ed62aa76588ac49017725/lib/kitchen/driver/ec2.rb#L266) doesn't get populated, so [update_username](https://github.com/test-kitchen/kitchen-ec2/blob/119001d1ad2e7879151ed62aa76588ac49017725/lib/kitchen/driver/ec2.rb#L295) which relies on it fails?

Possibly the same thing that @gretel's comment refers to [here](https://github.com/test-kitchen/kitchen-ec2/issues/225#issuecomment-199296835)

This PR works around it by not attempting to do anything with the username if actual_platform isn't defined (and therefore assumes you're setting the `transport.username` yourself).

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -287,9 +287,14 @@ module Kitchen
       end
 
       def update_username(state)
-        # TODO: if the user explicitly specified the transport's default username,
-        # do NOT overwrite it!
-        if instance.transport[:username] == instance.transport.class.defaults[:username]
+        # BUG: With the following equality condition on username, if the user specifies 'root'
+        # as the transport's username then we will overwrite that value with one from the standard
+        # platform definitions.  This seems difficult to handle here as the default username is
+        # provided by the underlying transport classes, and is often non-nil (eg; 'root'), leaving
+        # us no way to distinguish a user-set value from the transport's default.
+        # See https://github.com/test-kitchen/kitchen-ec2/pull/273
+        if actual_platform &&
+            instance.transport[:username] == instance.transport.class.defaults[:username]
           debug("No SSH username specified: using default username #{actual_platform.username} " \
                 " for image #{config[:image_id]}, which we detected as #{actual_platform}.")
           state[:username] = actual_platform.username
--- a/spec/kitchen/driver/ec2_spec.rb
+++ b/spec/kitchen/driver/ec2_spec.rb
@@ -382,6 +382,18 @@ describe Kitchen::Driver::Ec2 do
       include_examples "common create"
     end
 
+    context "instance is not a standard platform" do
+      let(:state) { {} }
+      before do
+        expect(driver).to receive(:actual_platform).and_return(nil)
+      end
+
+      it "doesn't set the state username" do
+        driver.update_username(state)
+        expect(state).to eq({})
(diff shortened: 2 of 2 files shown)


cheeseplus:
+1

lib/kitchen/driver/ec2.rb
```diff
@@ -289,7 +289,8 @@ def default_ami
       def update_username(state)
         # TODO: if the user explicitly specified the transport's default username,
         # do NOT overwrite it!
-        if instance.transport[:username] == instance.transport.class.defaults[:username]
+        if actual_platform &&
+            instance.transport[:username] == instance.transport.class.defaults[:username]
```
lamont-granquist: the bug is that if the user tried to set the username to the default where the username is not nil by default, then it'll go off and do the autodetection and may choose to override the default.  this check for username != default cannot catch that case -- unless the default value is something like nil which the user would never set.

it feels like someone was thinking too hard about default values here.

hynd: Neither of those `instance.transport`s are Nil for me (it's `actual_platform.username` that's coming up Nil when using non-"standard" AMI's, causing the error).

With a fresh setup and me not specifying anything, both of them are set to "root".  I can alter `instance.transport[:username]` by setting `transport.username` in my .kitchen.yml, and i think the default value for it - and `instance.transport.class.defaults[:username]` - are probably both coming from TK transport defaults? (ie; test-kitchen-1.13.0/lib/kitchen/transport/ssh.rb, which is "root")

I guess that's why that equality condition can't just check for `nil?` and the TODO still applies... (if i want to keep my TK username as "root" and i'm using the SSH transport, it looks like it'll always be overwritten by the `actual_platform.username` value?)

I didn't want to tinker with the existing username logic too much with just that comment for guidance, and if both of those `instance.transport` variables are getting defaults, i'm not sure how you'd test if the user was overriding one?  Maybe @jkeiser remembers a bit more about it?

lamont-granquist: okay it'd be good to get that information documented in the TODO so we don't have to re-discover it all again at some point in the future.  you might just rewrite the TODO more like "BUG:  the default is often non-nil (e.g. 'root') but if the user specifies 'root' explicitly then we will overwrite that value -- need a way to distinguish a default from a set value"

its amusing how every configuration system inevitably re-discovers this problem...

hynd: Reworded TODO to BUG with more info

robbkidd:
Two thumbs! Up for merge and gem release?

thomasrg:
Many thanks for merging this. Is there an ETA on the gem release?

PR #274, opened 2016-09-01
Add optional config for shutdown_behavior

alexpop:
This PR allows users to change the default [shutdown behavior](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#Using_ChangingInstanceInitiatedShutdownBehavior) of ec2 instances.

Example use-case: 
* Allow an instance to self-terminate if running for more than N hours.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,12 @@ The price you bid in order to submit a spot request. An additional step will be 
 
 The default is `nil`.
 
+### `instance_initiated_shutdown_behavior`
+
+Control whether an instance should `stop` or `terminate` when shutdown is initiated from the instance using an operating system command for system shutdown.
+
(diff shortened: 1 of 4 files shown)


tduffield:
According to the [docs](http://www.rubydoc.info/gems/aws-sdk-v1/1.64.0/AWS/EC2/Instance:instance_initiated_shutdown_behavior) the only valid options are "terminate" and "stop". Should we enforce that here as well?

alexpop:
Makes sense Tom, I'll make the change

marcparadise:
:+1:

rhass:
:+1:

mwrock:
:+1:

PR #283, opened 2016-10-17
Fix syntax in example

jbussdieker:
After some amount of testing it looks like the example usage of `image_search` should be placed under the `driver` key.


cheeseplus:
+1

PR #285, opened 2016-11-01
Add expiration to spot requests

alanbrent:
Currently, if a `kitchen converge` times out then the spot request does not get cleaned up. This means an operator has to manually remove the spot requests, leaving an opening for extra costs that are unnecessary and difficult to track down. 

It seems to me that we'd want the spot request to expire if it hasn't been fulfilled within the timeframe of `config[:retryable_tries] * config[:retryable_sleep]` seconds, so it would make sense to set an expiry on the request as such.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -356,9 +356,11 @@ module Kitchen
       end
 
       def create_spot_request
+        request_duration = config[:retryable_tries] * config[:retryable_sleep]
         request_data = {
           :spot_price => config[:spot_price].to_s,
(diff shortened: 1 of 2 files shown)


alanbrent:
Just wondering what needs to happen on my end to get some discussion on this PR?

cheeseplus:
We're actually trying to get a release of `kitchen-ec2` out so it was just a matter of running down what we can include. This LGTM and if we can get another reviewer this should be good to go.

thommay:
👍

alanbrent:
Awesome. Just let me know if anything is needed from me on this, otherwise I'll leave you alone :D

PR #291, opened 2017-02-06
Support Windows 2016

gdavison:
AWS has changed a number of characteristics of Windows Server 2016 instances:
- the AMI name pattern is now `Windows_Server-2016-English-<Style>-*`, where `<Style>` is one of `Full-Base` for regular servers, `Nano-Base` for Nano servers, or `Full-Containers` for server with containers
- The initialization process uses `EC2Launch` instead of `EC2Config`

http://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/windows-ami-version-history.html#win2k16-amis


Gemfile
```diff
@@ -8,7 +8,7 @@ gem "winrm-fs"
 gem "activesupport", "~> 4.0"
 
 group :test do
-  gem "rake"
+  gem "rake", "< 12"
```
cheeseplus: Any reason for this particular pin?

.travis.yml
```diff
@@ -2,6 +2,5 @@ language: ruby
 cache: bundler
 sudo: false
 rvm:
-  - 2.0.0
-  - 2.1
-  - 2.2
+  - 2.1.8
```
cheeseplus: Safe to drop 2.1.x and keep somewhat in line with https://github.com/test-kitchen/test-kitchen/blob/master/.travis.yml#L3-L6

cheeseplus:
Aside from the inline comments this look great!

Gemfile
gdavison: Rake 12 isn't compatible with rubocop 0.28, which is pinned by finstyle 1.4.0.

And the gemspec file says 

> style and complexity libraries are tightly version pinned as newer releases may introduce new and undesireable style choices which would be immediately enforced in CI

gdavison: I just tried updating to finstyle 1.5.0. It still isn't compatible with rake 12, and it picks up 11 more style offences :)

cheeseplus: We're moving away from finstyle and I didn't realize this repo was still on it so new things to do

PR #305, opened 2017-03-01
Wrap create and destroy actions in retryable blocks

relu:
I've recently had a lot of issues with `Request limit exceeded.` exceptions being thrown due to high usage. I've been googling and stumbled upon this comment which pretty much suggested the solution I came up with here: https://github.com/test-kitchen/kitchen-ec2/pull/44#issuecomment-278463973

I'm submitting my proposed changes here in the hopes this might come in handy for others.


cheeseplus:
Two questions:
- could you add tests?
- should we expose the number of retries as toggleable thing?

relu:
@cheeseplus added the changes you proposed.

I think it's a good point to have the number of retries configurable, thus I've added `retry_on_aws_error_tries` which defaults to `10`.

coderanger:
So making create retry automatically needs some more careful logic I think. We need to be relatively certain we aren't leaking 10 machines each time just because something is timing out.

relu:
Maybe a more granular approach is needed, where we'd wrap each of the api caller methods in a retryable block. Or checking the server variable prior to making another `submit_spot` or `submit_server` call. Will follow-up.

coderanger:
Like barring a CAP theorem break we can't perfectly guarantee that we'll never double-create or miss a destroy in the case of byzantine failures, but something this simple seems likely to do more harm than good often :)

relu:
I'd like to emphasize I'm only targeting a particular exception which is `Aws::EC2::Errors::RequestLimitExceeded`, has nothing to do with timeouts.
In case of a request limit exceeded exception that can happen at any point (even on destroy) the instance(s) will still dangle around until manually terminated. At least that's my experience with this issue.
My solution here covers for exactly this case.

hfinucane:
@coderanger [ec2.client.run_instances](http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html#run_instances-instance_method) takes a `:client_token` that could be used to prevent double-creates. That said, given that it only retries on AWS API errors, not network errors, I'm not sure there would be a real difference, unless you're asserting that Aws::EC2::Errors::RequestLimitExceeded or Aws::EC2::Errors::InvalidInstanceIDNotFound can be returned even though the operation succeeds.

I run kitchen-ec2 fairly heavily, and my solution to `destroy`s that don't happen because of some sort of crash or network error is to have a separate cleanup job that just looks for instances over a certain age. Given the full universe of possible failures on the test runner's side, I suspect everyone is going to have to give up and write that job, too.

I am super interested in this patch. Currently I have a setup that rate limits test-kitchen to something like 8 concurrent tests per cookbook, and rate limit my CI to one test-kitchen run at a time, and *still* occasionally see spurious integration testing failures because some other, non-test-kitchen process is trying to use EC2. This seems like a much nicer, and much less coarse-grained solution.

lib/kitchen/driver/ec2.rb
```diff
@@ -230,21 +243,24 @@ def create(state) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
...
-          server.terminate
-        end
-        if state[:spot_request_id]
-          debug("Deleting spot request <#{state[:server_id]}>")
-          ec2.client.cancel_spot_instance_requests(
-            :spot_instance_request_ids => [state[:spot_request_id]]
-          )
-          state.delete(:spot_request_id)
+        retry_on_aws_error do |r|
+          info("Attepmpting to destroy instance <#{state[:server_id]}>, #{r} retries")
```
hfinucane: s/Attepmpting/Attempting

pantocrator27:
just out of curiosity where are we on this?  this seems cool 👍

pantocrator27:
Just a thought here ... AWS typically recommends exponential backoff retries: http://docs.aws.amazon.com/general/latest/gr/api-retries.html.  Would it be worthwhile to not only specify the number of retries to do, but the length of time between retries and then the retry would be done in the exponential backoff retry fashion?

pantocrator27:
I think I spoke too soon ... I see it ... would it still be worth specifying time of retries?

relu:
@pantocrator27 wouldn't adding that option defeat the purpose of having exponential backoff retries, I mean wouldn't this basically set a fixed interval between retries? This would go against the recommendation, so I'm not sure if it makes sense to add it since it can cause unexpected results. Or maybe I misunderstood your proposal...

pantocrator27:
@relu sorry i was not clear.  typically when I have written this type of logic i have seeded this type of logic with an initial time for a retry which it looks like you have you have set at `30`.  Do we want to make this configurable?  Additionally at what point do we stop retrying?  Do we want to enforce either a time limit or a number of retries?

relu:
@pantocrator27 got it, you're referring to this
```
lambda { |n| [2**n, 30].min },
```
This is not a seed value, it's the maximum amount of time we allow Retryable to sleep. If `2**n` goes above 30, we use 30.
It's kind of like this after each retry: 2, 4, 8, 16, 30, 30, 30, 30, 30, 30 seconds (with 10 retries).

I think in this context it might not make sense to have a higher value, waiting one full minute between retries has no added benefit over retrying every 30 seconds if it gets to that). The idea is to not prolong the time it takes to complete a certain call more than it is necessary.
I can see the utility of having this value configurable, for instance if one might consider doing more or less aggressive retrying, however setting this value too low can potentially cause the action to ultimately fail since we can run out of retries very quickly, while setting it too high can cause significant delays in execution (which might not be desired).

Overall I think setting 30 seconds as the maximum wait interval should be good enough, don't really think it requires tweaking in most cases. On very busy AWS accounts one might just want to increase the number of retries...

pantocrator27:
My apologies.  I misread the code.  Please ignore me :)

pantocrator27:
👍

pantocrator27:
Maybe this is an idea for a separate PR but I am also getting a lot of `Failed to complete #create action: [We currently do not have sufficient m4.large capacity in the Availability Zone you requested (us-east-1b). Our system will be working on provisioning additional capacity. You can currently get m4.large capacity by not specifying an Availability Zone in your request or choosing us-east-1a, us-east-1c, us-east-1d, us-east-1e.]` lately and wondered if the same retry logic is worth persuing

optix2000:
After using your patch for a while, I still occasionally get "Request limit exceeded" errors, it comes from `Aws::Waiters` instead of `Aws::EC2`.

`Aws::Waiters::Errors::UnexpectedError: stopped waiting due to an unexpected error: Request limit exceeded.`

Not sure if this is a bug on the AWS-SDK, or another edge case to handle for hitting request limits.

Backtrace
```
Kitchen::ActionFailed: Failed to complete #create action: [stopped waiting due to an unexpected error: Request limit exceeded.]
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:111:in `block in poll'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:101:in `loop'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:101:in `poll'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:91:in `block (2 levels) in wait'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:90:in `catch'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:90:in `block in wait'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:89:in `catch'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:89:in `wait'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/client_waiters.rb:110:in `wait_until'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-resources-2.9.16/lib/aws-sdk-resources/operations.rb:229:in `call'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-resources-2.9.16/lib/aws-sdk-resources/operation_methods.rb:19:in `block in add_operation'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:202:in `block in create'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:635:in `block in retry_on_aws_error'
vendor/bundle/ruby/2.2.0/gems/retryable-2.0.2/lib/retryable.rb:67:in `retryable'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:630:in `retry_on_aws_error'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:201:in `create'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:460:in `public_send'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:460:in `block in perform_action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:527:in `synchronize_or_call'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:489:in `block in action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:488:in `action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:460:in `perform_action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:370:in `create_action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:359:in `block in transition_to'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:358:in `each'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:358:in `transition_to'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:157:in `verify'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:186:in `block in test'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:182:in `test'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/rake_tasks.rb:59:in `block (3 levels) in define'
vendor/bundle/ruby/2.2.0/gems/rake-11.3.0/exe/rake:27:in `<top (required)>'
Aws::Waiters::Errors::UnexpectedError: stopped waiting due to an unexpected error: Request limit exceeded.
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:111:in `block in poll'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:101:in `loop'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:101:in `poll'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:91:in `block (2 levels) in wait'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:90:in `catch'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:90:in `block in wait'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:89:in `catch'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/waiters/waiter.rb:89:in `wait'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.9.16/lib/aws-sdk-core/client_waiters.rb:110:in `wait_until'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-resources-2.9.16/lib/aws-sdk-resources/operations.rb:229:in `call'
vendor/bundle/ruby/2.2.0/gems/aws-sdk-resources-2.9.16/lib/aws-sdk-resources/operation_methods.rb:19:in `block in add_operation'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:202:in `block in create'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:635:in `block in retry_on_aws_error'
vendor/bundle/ruby/2.2.0/gems/retryable-2.0.2/lib/retryable.rb:67:in `retryable'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:630:in `retry_on_aws_error'
vendor/bundle/ruby/2.2.0/gems/kitchen-ec2-1.3.2/lib/kitchen/driver/ec2.rb:201:in `create'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:460:in `public_send'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:460:in `block in perform_action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:527:in `synchronize_or_call'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:489:in `block in action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:488:in `action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:460:in `perform_action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:370:in `create_action'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:359:in `block in transition_to'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:358:in `each'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:358:in `transition_to'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:157:in `verify'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:186:in `block in test'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/instance.rb:182:in `test'
vendor/bundle/ruby/2.2.0/gems/test-kitchen-1.16.0/lib/kitchen/rake_tasks.rb:59:in `block (3 levels) in define'
vendor/bundle/ruby/2.2.0/gems/rake-11.3.0/exe/rake:27:in `<top (required)>'```

relu:
We could catch this `Aws::Waiters::Errors::UnexpectedError` exception as well in the retry block, but I'm not sure if this is the right solution.

Good catch @optix2000, didn't consider the waiters would also cause request limit exceptions. Unfortunately the exception thrown is pretty genertic, it would have been easier if a similar exception to the one ec2 throws `RequestLimitExceeded` would have been thrown by the waiters.

Retryable also has a `matching` argument, which can take a string or regular expression. Might be considered if it works in conjunction to the `on` argument we're already using. Will look into it.

relu:
@optix2000 I've updated my branch with some changes that would cover for the waiter exception on all occasions. Please check it out and see if it solves the problem. It works well for me, however I was not able to actually reproduce your described case directly.

optix2000:
Thanks for the quick fix and the great patch! (Solves a lot of issues with running kitchen tests in parallel on EC2)

I'll try it out and see if I still run into it. Will let you know in a week or so if I see it pop up again.

ChadScott:
How is this different than the existing retry logic within the AWS SDK Gem?

https://github.com/test-kitchen/kitchen-ec2/blob/aac18fd2f80a0522d658c12cf4baeebbeb62fe50/lib/kitchen/driver/aws/client.rb#L54

relu:
@ChadScott I think what you're saying is right, the scope of this PR needs to be re-evaluated. I remember having some issues with the default retry behavior of the aws-sdk gem (one issue is documented [here](https://github.com/test-kitchen/kitchen-ec2/blob/master/lib/kitchen/driver/ec2.rb#L195), though it might not be relevant to the context of these changes).
Since `retry_limit` is configurable, might be worth testing some more and adjusting that value rather than using this solution.

cheeseplus:
Closing this one in favor of the imp in #320

PR #307, opened 2017-04-01
Atomic tags

freimer:
Use new AWS functionality to tag both the instance and volumes at instance creation (RunInstances API call), rather than after the instance is created.

Requires updated AWS Ruby SDK.  I don't know how to enforce that.  Perhaps modify kitchen-ec2.gemspec to set the gem.add_dependency from `"aws-sdk", "~> 2"` to `"aws-sdk", "> 2.9.1"`???  I'm sure you know how to do that ;-)

### DIFF
--- a/kitchen-ec2.gemspec
+++ b/kitchen-ec2.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "test-kitchen", "~> 1.4", ">= 1.4.1"
   gem.add_dependency "excon"
   gem.add_dependency "multi_json"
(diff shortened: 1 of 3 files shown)


cheeseplus:
We were looking at merging #335 but this PR pre-dates and wanted to give a chance to rebase this against the latest work. I think we'll still need the volume tagging logic potentially because of the issues we'd seen in the code comments but curious if this is something you saw in testing.

cheeseplus:
Closing this one given https://github.com/test-kitchen/kitchen-ec2/pull/364

PR #320, opened 2017-06-12
Use quadratic backoff when encountering RequestLimit errors

kamaradclimber:
fetch_windows_admin_password can be subject to RequestLimit errors as
well

Signed-off-by: Grégoire Seux <g.seux@criteo.com>


cheeseplus:
Many thanks as always.

PR #321, opened 2017-06-20
Add support for Amazon Linux

steven-burns:


troyready:
I haven't done this with the ruby SDK, but here are some notes based on my experience with boto and their image publishing:

1) It's probably overkill, but I like including the following filters to make sure only relevant AMIs are returned:
```
[
        {
            'Name': 'name',
            'Values': ['amzn-ami-hvm*x86_64-gp2']
        },
        {
            'Name': 'architecture',
            'Values': ['x86_64']
        },
        {
            'Name': 'root-device-type',
            'Values': ['ebs']
        },
        {
            'Name': 'state',
            'Values': ['available']
        },
        {
            'Name': 'virtualization-type',
            'Values': ['hvm']
        }
]
```
Need to swap in the specific version in the name, obviously. Not sure if this is too opinionated a search, but I think it might be right.

2) When I sort the list of images, I'll sometimes end up with RC versions (e.g. `amzn-ami-hvm-2017.03.rc-1.20170327-x86_64-gp2` appearing before the actual latest release(e.g. `amzn-ami-hvm-2017.03.1.20170617-x86_64-gp2`). Not sure if this is a specific to python sorting, but it might be worth explicitly excluding them here too.

3) When I do boto3.client('ec2').describe_images(), I can specify `Owners=['amazon']`, which seems a little nicer to me than a specific owner-id. Not sure if that's applicable here or even worth mentioning.

Hope this helps; would love to see this functionality included!

Issue #322, opened 2017-06-22
Windows AMIs that don't have word 'windows' in image.name result in undefined methodversion' for nil:NilClass


vinyar:
The Windows2016 AMI we're using does not contain the word 'windows' in the AMI name.
This results in an error: `[undefined method `version' for nil:NilClass`

The error seems to come from this code:
https://github.com/test-kitchen/kitchen-ec2/blob/master/lib/kitchen/driver/aws/standard_platform/windows.rb#L54

Full debug gist:
https://gist.github.com/mbenachour/005aa718b8abb2978d9f6cb01e2f0644

@mbenachour

vinyar:
We've cloned the AMI and renamed it to contain 'windows' in the AMI name and that got us moving forward.

BenLiyanage:
Just posted a PR that fixes this issue--if you can still reproduce this do you mind testing out my branch?

jclynny:
Would love to see some traction on this, have been battling this for a bit, now I've put in the suggested workaround to have 'windows' in the AMI name, but I'd really like to see this work as it has in prior platform versions, and Linux, I don't believe I've ever had to specify the platform name in the AMI Name prior.

PR #324, opened 2017-06-29
removed actual_platform dependency when picking log location

BenLiyanage:
This PR fixes the issue here.  I was having the same issue.

https://github.com/test-kitchen/kitchen-ec2/issues/322

Included in this PR:
* Refactored powershell script to be in order in the code.
* Removed the `actual_platform.version` check which was throwing the error when using a custom AMI.
* Set log file to a consistent location that is always present when on a windows server.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -535,10 +535,28 @@ module Kitchen
       end
 
       def default_windows_user_data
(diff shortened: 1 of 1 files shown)


cheeseplus:
We've got a merge conflict here but I'm happy to get this in with a rebase. I had also merged a bunch of other Windows (2016) related fixes recently so would like to squash this one as well if it's still needed.

tas50:
Merged in #387

PR #329, opened 2017-07-17
Pass named profile to Aws::Config directly

BrentOnRails:
Kitchen-ec2 should allow aws-sdk to decide best credentials when passed a properly configured profile.

Aws::SharedCredentials does not negotiate some profile features that are available.  (some attempts to patch the module in the aws-sdk-ruby repo have gone stale)

Example (assume role):
```
[auth]
aws_access_key_id=foobar
aws_secret_access_key=foobarbaz

[use1dev]
source_profile = auth
region = us-east-1
role_arn = arn:aws:iam::12345678910:role/auth-admin
```

I'm proposing to pass the profile directly to the config/client when user-specified, rather than passing it through SharedCredentials, which does not support some credentials file functionality.

Edit:  Specifically this brings kitchen-ec2 in line with aws-cli functionality described here http://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html

### DIFF
--- a/lib/kitchen/driver/aws/client.rb
+++ b/lib/kitchen/driver/aws/client.rb
@@ -42,15 +42,22 @@ module Kitchen
           retry_limit = nil,
           ssl_verify_peer = true
         )
-          creds = self.class.get_credentials(
-            profile_name, access_key_id, secret_access_key, session_token, region
-          )
+
           ::Aws.config.update(
             :region => region,
-            :credentials => creds,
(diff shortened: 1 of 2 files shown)


avanier:
Oooooh! Shiny! I've been humped by this on multiple occasions!

Being able to legitimately use AWS roles would be great!

I strongly approve of this PR! :stuck_out_tongue:

cheeseplus:
Closing in favor of #353

Issue #333, opened 2017-07-27
ECS task role ignored


llibicpep:
I am unsure this is kitchen-ec2 issue, but accordingly to https://github.com/aws/aws-sdk-ruby/issues/1309 generally it should work. I have validated that my container has `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` set and also that AWS CLI picking it up correctly, so I can only guess something happens within kitchen-ec2.
I've tested on 1.3.2 and 1.2.0 in combination with a few ruby aws-sdk recent versions, and current behaviour is that it's picking up cluster EC2 role instead of task role.

llibicpep:
Looks like the issue is here https://github.com/test-kitchen/kitchen-ec2/blob/v1.3.2/lib/kitchen/driver/aws/client.rb#L77. It's picking up `InstanceProfileCredentials` as last resort not considering `ECSCredentials` and `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` and not using credentials chain to delegate this to sdk.

cheeseplus:
This could be added as a new feature, probably just missed as most folks aren't using kitchen-ec2 with ECS.

cheeseplus:
We've cut kitchen-ec2 2.0.0 which basically dumps all our custom logic and defers to the AWS SDK so this should be resolved once that gem is updated in all the appropriate places.

llibicpep:
@cheeseplus I just tested with kitchen-ec2 2.0.0 - it is now correctly picking up task role instead of cluster instances role. Thanks!

PR #334, opened 2017-07-27
#333: Use ruby aws-sdk CredentialProviderChain to resolve credentials

llibicpep:


robbkidd:
😬  I am wary of relying on [a private API object](https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb#L2) for resolving credentials.

llibicpep:
Is it better to creating multiple conflicting sources of truth? I have no idea why this is private in ruby sdk, while this is [suggested, documented](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default) and [public](https://github.com/aws/aws-sdk-java/blob/7844c64cf248aed889811bf2e871ad6b276a89ca/aws-java-sdk-core/src/main/java/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.java#L44) way of doing same thing in other platforms.

cheeseplus:
Closing in favor of #353

PR #335, opened 2017-08-29
Tag specifications

deepsihag:


cheeseplus:
We already support this via https://github.com/test-kitchen/kitchen-ec2#tags so I'm not sure what this PR is doing other than re-implementing tags again.

deepsihag:
`tags` are added in a separate AWS call than instance creation. In case of a restricted policy where tags should be applied at the time of machine creation, [tag_specifications ](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_TagSpecification.html) needs to be defined. So this PR is just an extension to support that.

cheeseplus:
Ok, so I think I see where the difference is now but my hesitation here is adding _another_ and potentially conflicting option. I think the better route here is to fix the existing tags implementation a la https://github.com/test-kitchen/kitchen-ec2/pull/307 instead of having two places to set them.

Does that make sense? I think could enhance this PR to support that and that'd be ideal.

patkar:
:-1: for this PR
:+1: for #307, so tags specified on the RunInstance call, w/o renaming the `tags` key in the test-kitchne ec2 driver config

cheeseplus:
Closing for the reasons outlined above and new pending PR https://github.com/test-kitchen/kitchen-ec2/pull/364

PR #338, opened 2017-09-19
Handle nulls/binary text in user data so it supports gzip

brodygov:
`File.file?("\0")` raises ArgumentError, so it was previously impossible
to set binary data (such as gzipped data) in EC2 user_data.

This PR also adds a test that would have caught the bug.


cheeseplus:
Amazing catch <3

PR #339, opened 2017-09-21
Update AWS-SDK to v3

tristanmorgan:
This PR updates the Gem to use the aws-sdk-iam and aws-sdk-ec2 gems rather than load the whole SDK.

### DIFF
--- a/kitchen-ec2.gemspec
+++ b/kitchen-ec2.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "test-kitchen", "~> 1.4", ">= 1.4.1"
   gem.add_dependency "excon"
   gem.add_dependency "multi_json"
-  gem.add_dependency "aws-sdk", "~> 2"
+  gem.add_dependency "aws-sdk-ec2", "~> 1.42"
   gem.add_dependency "retryable", "~> 2.0"
 
   gem.add_development_dependency "rspec",     "~> 3.2"
--- a/lib/kitchen/driver/aws/client.rb
+++ b/lib/kitchen/driver/aws/client.rb
@@ -17,7 +17,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "aws-sdk"
+require "aws-sdk-ec2"
 require "aws-sdk-core/credentials"
 require "aws-sdk-core/shared_credentials"
 require "aws-sdk-core/instance_profile_credentials"
@@ -50,7 +50,7 @@ module Kitchen
         end
 
         def create_instance(options)
-          resource.create_instances(options)[0]
+          resource.create_instances(options).first
         end
 
         def get_instance(id)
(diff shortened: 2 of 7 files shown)


tristanmorgan:
I've hit a deprecation warning. Working on it still.

lamont-granquist:
looks like aws-sdk-v3 conflicts strictly with aws-sdk-v2 and cannot be bundled in the same gemfile.  that's going to block merging this until chef-provisioning-aws also gets merged:

```
Bundler could not find compatible versions for gem "aws-sdk-core":
  In Gemfile:
    kitchen-ec2 was resolved to 1.3.2, which depends on
      aws-sdk-ec2 was resolved to 1.0.0, which depends on
        aws-sdk-core (~> 3)

    chef-provisioning-aws (>= 2.0) was resolved to 2.0.0, which depends on
      aws-sdk (< 3.0, >= 2.1.26) was resolved to 2.1.26, which depends on
        aws-sdk-resources (= 2.1.26) was resolved to 2.1.26, which depends on
          aws-sdk-core (= 2.1.26)
```

there's also a bit of a concern that i've got which is that this is only the conflict that occurs in the gems that ship in the ChefDK by default, and the whole ecosystem looks like its going to need to take a flag day to upgrade to aws-sdk v3, unless i've missed something here.

tristanmorgan:
OK, well until ChefDK gets the version bump we know its a simple change to update.

arothian:
@lamont-granquist Any idea on when this might get done? Running into issues bundling with other gems that have migrated to sdk 3.

tristanmorgan:
It seems Chef/Train also introduced a dependancy on the old V2 SDK too. https://github.com/chef/train/pull/239

kitchen-ec2.gemspec
```diff
@@ -23,7 +23,8 @@ Gem::Specification.new do |gem|
   gem.add_dependency "test-kitchen", "~> 1.4", ">= 1.4.1"
   gem.add_dependency "excon"
   gem.add_dependency "multi_json"
-  gem.add_dependency "aws-sdk", "~> 2"
+  gem.add_dependency "aws-sdk-iam"
```
grosser: prefer to depend on a major version like `~> 1.2` so we don't bump across major versions would testing them

grosser:
is there any issue open to get ChefDK updated ?

coderanger:
@grosser Not currently, there is not currently a maintainer for chef-provisioning-aws which complicates things.

tristanmorgan:
Unfortunately chef/train also introduced a dependancy on the old AWS-SDK v2.

tas50:
This is very much something we want to ship so we can slim down the size of ChefDK, but there's additional work outside of kitchen-ec2 we have to do:

- Update chef-provisioning-aws
- Update InSpec

It's going to happen, but I don't have a timeline.

tas50:
Closing this out in favor of #419

Issue #318, opened 2017-05-26
kitchen-ec2 only works with instances with provisioned EBS volumes


wolftrouble:
With newer versions of kitchen-ec2 the function wait_until_volumes_ready is always called on create(); and in the most recent version (https://github.com/test-kitchen/kitchen-ec2/blob/master/lib/kitchen/driver/ec2.rb#L409) is the following code"

`(described_volume_count > 0) && (described_volume_count == ready_volume_count)`

The problem here is if no EBS volumes are configured (e.g. only instance store is used) this function will never succeed, because described_volume_count == 0 in those cases. 

I can't find any evidence this is by design, and it causes problems for us as we do use some instance-only (no EBS) storage.  Is this intentional? If not, I'd think the test for described_volume_count > 0 should be simply removed entirely - a server with no described volumes need not wait on volumes being ready to proceed.

cheeseplus:
Fixed by #343

wolftrouble:
Unless I'm mistaken, this fix also has a pretty serious bug. Now we have the opposite problem: if using kitchen-ec2 with a non-root EBS volume it introduces a race condition, because now the instance setup won't wait on other, non-root provisioned EBS volumes to finish setup before considering the server ready. 

I would think the safe logic would be: "If AMI is root-EBS or if provisioned_voumes > 0 then wait for volumes, otherwise do not wait for EBS."

PR #343, opened 2017-10-08
Fix for issue with instance-store backed instance (issue #318)

naunga:
Added logic to ensure that the driver will not attempt to tag volumes or wait for volumes to be ready when the AMI being used has a rootDeviceType of "instance-store" instead of "ebs".


cheeseplus:
This is great!

Issue #331, opened 2017-07-26
Windows 2016 boxes not being seen as 'ready' using ec2 driver


joshoohaah:
I have kitchen yml file that deploys 2012 boxes successfully, however when i try 2016 boxes they wait for it to become ready until the process times out.  I have it set for trying 180 times (900 seconds) which is more than enough time.  Windows 2012 boxes typically come on around 160 seconds.
Does 2016 need to do something different than 2012 to return a ready flag back to kitchen-ec2 driver? 

kitchen yml file for reference:
<script src="https://gist.github.com/joshoohaah/4a37d843bd80ff70dfd0492e8cf0b2ee.js"></script>

Out of relevant kitchen run in -l debug before it times out eventually:
```
       Instance <i-020a3b0f4c64d06da> requested.
       Polling AWS for existence, attempt 0...
       Attempting to tag the instance, 0 retries
       EC2 instance <i-020a3b0f4c64d06da> created.
       Waited 0/900s for instance <i-020a3b0f4c64d06da> volumes to be ready.
       Waited 0/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 5/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 10/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 15/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 20/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 25/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 30/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 35/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 40/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 45/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 50/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 55/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 60/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 65/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 70/900s for instance <i-020a3b0f4c64d06da> to become ready.
       Waited 75/900s for instance <i-020a3b0f4c64d06da> to become ready.
[...]

Issue #319, opened 2017-06-09
Windows 2016 Fails to Connect When Using Custom Credentials


jdgoins:
When I provide custom credentials via my .kitchen.yml file, the ec2 driver is unable to connect to a new w2k16 instance with the supplied credentials. I am using w2k16 base latest (ami-c2c3a2a2 as of this writing). I have confirmed that the custom user is created and added to the Administrators group. This is not an issue on w2k12r2.

The run gets stuck on `[WinRM] opening remote shell on http://ec2-xx-xx-xx-xx.us-west-2.compute.amazonaws.com:5985/wsman`

The driver is able to connect using the Administrator account when I remove this portion from my .kitchen.yml file, transport section.
```
  username: chef
  password: '3rF5vBqnMEwmmWs4'
```

grdnrio:
I also have this issue.

Looking at the code you can see the user-data being used to enable winrm access, but it doesn't seem to work on the latest 2016 AMI. I haven't tried running this manually, however. https://github.com/test-kitchen/kitchen-ec2/blob/master/lib/kitchen/driver/ec2.rb#L537-L585

It similar to what happens in the kitchen-azure driver: https://github.com/test-kitchen/kitchen-azurerm/blob/master/lib/kitchen/driver/azurerm.rb#L333-L379.

I'd assume that using the `user_data` property would override this, so it's probably a case of setting this for the 2016 platform. https://github.com/test-kitchen/kitchen-ec2#user_data

I'll share my user-data if I get it working.

pschaumburg:
Can confirm the issue on my site.

joshoohaah:
with w2k16 i am having just issues with kitchen-ec2 giving information back that the box is running.  In the AWS console it shows up and running. but in the debug output for test-kitchen it just keeps waiting for the box to become ready.
```
       Waited 890/900s for instance <instance_ID> to become ready.
D      Console output: ---
2017/07/26 16:00:17Z: Amazon SSM Agent v2.0.847.0 is running
```

Sam-Martin:
I was having a similar issue, not sure if it would have worked had I removed the custom password. BUT I was able to resolve it by adding:

```
Set-ItemProperty -Name LocalAccountTokenFilterPolicy -Path HKLM:\software\Microsoft\Windows\CurrentVersion\Policies\system -Value 1
```
[...]

PR #348, opened 2017-11-01
Set LocalAccountTokenFilterPolicy to allow powershell remoting from local accounts

Sam-Martin:
LocalAccountTokenFilterPolicy **should** be set by `Enable-PSRemoting`, but doesn't seem to be on 2016. 
I had issues similar to those presented in #319 and #331, this will should fix both.


cheeseplus:
Ace!

Issue #344, opened 2017-10-12
Support for STS assumeRole


patkar:
In our company we're solely using only one credentials set for a master account, were we can assumeRole into any other AWS account we manage. Would be good to have an assume_role parameter (with role ARN), so this driver will first obtain a new access credentials set from STS for the assumed role.

patkar:
Additionaly it could extend to support the `role_arn` from `~/.aws/config`:

```
[profile client-project]
role_arn = arn:aws:iam::1122234567890:role/Developer
source_profile = pixelart
```

rhyas:
@patkar Check out the PR I just submitted. We're using that internally to do exactly what this ticket is asking for, making use of the config file for source_profile and role_arn. All you need in the kitchen.yml is the name of the profile you're aiming to use, the SDK handles the rest.

patkar:
Ok, your PR means, it is necessary to correctly configure every profile we want to assume in `~/.aws/config`. It's ok, at least something which forces us to manage that cleanly ^^

rhyas:
Yes, previously the config file wasn't really used. Now it is, so if it's being used, it has to be formatted right. However, you still have all the other options as well, such as only using the credentials file, or using machine roles, environment variables, etc. This PR just moves all that detection back to the SDK and enables support for modes the SDK supports but the module didn't because it did discovery on it's own.

patkar:
Yeah, forcing users to configure `~/.aws/config` correctly is a legit approach. IMHO every other application can profit from it too. Although I still see many not supporting it :)

Finally it's a better approach, then having a role_arn parameter and handling it itself. (Side note: Terraform AWS provider supports roles from `~/.aws/config` in the newest release :grin: Reason why I approached a `role_arn` param first, 'cause I was used to it through Terraform ^^).

cheeseplus:
PR merged and 2.0.0 released, PR for ChefDK is incoming so this should be available in a `current` channel build soon.

PR #353, opened 2017-11-20
Clean up original Authentication; Rely on SDK for Chain.

rhyas:
This PR removes the logic for searching for authentication credentials in favor of just relying on the SDK to find credentials through the standard means. This is necessary to allow role delegation using source_profile configurations for credentials. None of the authentication classes the SDK provides will support that, but if you let the SDK just do all the work to find the credentials, everything just works. This also greatly simplifies the logic within the kitchen-ec2 client wrapper.

I suspect this will break most/all of the tests in client_spec.rb because everything it tests is gone. I vote for just removing it, and passing the buck to the SDK for both finding, and testing all the various methods, of credential finding. I didn't remove the tests in this PR yet, pending discussion on that point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,35 +147,18 @@ follow the other `image_search` rules for preference.
 
 ### AWS Authentication
 
-In order to connect to AWS, you must specify the AWS access key id and secret key
-for your account. There are 3 ways you do this, and we will try them in the
-following order:
+In order to connect to AWS, you must specify AWS credentials. We rely on the SDK
+to find credentials in the standard way, documented here: 
+https://github.com/aws/aws-sdk-ruby/#configuration
 
-1. You can specify the access key and access secret (and optionally the session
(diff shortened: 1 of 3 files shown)


rhyas:
This will resolve #344.

rhyas:
Cleaned up the test suite to remove the creds testing, since it doesn't exist anymore.

cheeseplus:
This looks great! We'll give this a look but the simplification is long overdue.

rhyas:
This was tagged for 1.4.0, but didn't make it into the release. Is there a reason? And is there a plan to get it merged anytime soon? Trying to use this across a team, and maintaining a one-off is tedious. (:

cheeseplus:
It was tagged as 1.4. for a hot minute before we decided it needed to be a major release thing and there were lots of other fixes that could go out with less impact. I'm gearing up for the 2.x release (granted this is the most major change) shortly but 1.4 has only been out a day or so.

PR #362, opened 2017-12-15
Automatically create a security group and key pair if needed.

coderanger:
Still needs docs, but code review can start. If a security group or key pair name isn't given, it makes one for you.

Also adds an option to skip the cost warning and cleans up some defunct arguments in the Client class.


coderanger:
I tried to deal with the various ways we could leak them, but it's hard to cover every weird case. This does make `destroy` a bit slower if you are using this feature since the instance needs to actually terminate between we can remove the SG.

coderanger:
Should be ready to go :)

Issue #380, opened 2018-02-02
Nil tags cause the provisioning to fail


tas50:
If the kitchen config accidentally includes nil values for a tag then the kitchen provisioning will fail in a way that leaves behind a node which must be manually deleted.

With the following tags that use ENV values, but the ENV values were not set:

```yaml
platforms:
  - name: ubuntu-16.04
    driver:
      tags:
        - commit: <% ENV['TRAVIS_COMMIT'] %>
        - pr: <% ENV['TRAVIS_PULL_REQUEST'] %>
```

```
       Instance <i-0b54ac25c7e283ef5> requested.
       Polling AWS for existence, attempt 0...
       Attempting to tag the instance, 0 retries
       Removing automatic security group sg-82d403fd
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #create action: [resource sg-82d403fd has a dependent object] on base-ubuntu-1604
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

When the tags have been commented out:

```
       Instance <i-0aad1be64b56169ad> requested.
       Polling AWS for existence, attempt 0...
       Attempting to tag the instance, 0 retries
       EC2 instance <i-0aad1be64b56169ad> created.
       Waited 0/300s for instance <i-0aad1be64b56169ad> volumes to be ready.
       Waited 0/300s for instance <i-0aad1be64b56169ad> to become ready.
       Waited 5/300s for instance <i-0aad1be64b56169ad> to become ready.
       Waited 10/300s for instance <i-0aad1be64b56169ad> to become ready.
       Waited 15/300s for instance <i-0aad1be64b56169ad> to become ready.
       Waited 20/300s for instance <i-0aad1be64b56169ad> to become ready.
       EC2 instance <i-0aad1be64b56169ad> ready (hostname: 10.0.0.41).
       Waiting for SSH service on 10.0.0.41:22, retrying in 3 seconds
       Waiting for SSH service on 10.0.0.41:22, retrying in 3 seconds
       Waiting for SSH service on 10.0.0.41:22, retrying in 3 seconds
       Waiting for SSH service on 10.0.0.41:22, retrying in 3 seconds
       Waiting for SSH service on 10.0.0.41:22, retrying in 3 seconds
[...]

PR #381, opened 2018-02-02
Fix handling of nil/Integer tags and add logging if improper format is used

tas50:
Previous bugs:

- Tag values in AWS can't be Integers, but YAML makes it easy to pass an integer so lets convert those to Strings.
- If you didn't read the example you would probably think 1 tag per line which is incorrect. Lets throw a warning when we see an array instead of a hash, which indicates someone did this.
- If you want to pass a nil value tag you have to do it as an empty string so we need to convert Nils to Strings too.

Signed-off-by: Tim Smith <tsmith@chef.io>

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -453,7 +453,7 @@ module Kitchen
       def tag_server(server)
         if config[:tags] && !config[:tags].empty?
           tags = config[:tags].map do |k, v|
-            { :key => k, :value => v }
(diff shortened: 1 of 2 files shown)


tas50:
Fixes https://github.com/test-kitchen/kitchen-ec2/issues/380

tas50:
@cheeseplus This should be good to go now. It took me a while to track down my random problems, but in the end we had some formatting issues in how tagging worked. This should save the next person a pile of time.

PR #382, opened 2018-02-03
Allow 'host' type tenancy & add proper validation warnings

tas50:
Do this in a validation warning so someone actually knows what the value should be. Right now we'd just silently skip it if someone provided host, which is a pretty bad experience.

Signed-off-by: Tim Smith <tsmith@chef.io>


tas50:
@cheeseplus This one should be good to go as well

PR #385, opened 2018-02-08
Don't create security group if security_group_filter is set

dpattmann:
If we specify a `security_group_filter` and no `security_group_id` the filter will be ignored and a default `security_group` will be created. This PR should fix this issue.


cheeseplus:
Thanks for the PR! The build failure is erroneous right now but first pass this looks good to me.

PR #392, opened 2018-03-14
allow AWS-managed ssh key pairs to be disabled

### LINKED ISSUE #391: Explicitly support usage w/o manual or autoconfiguration of aws_ssh_key_id
I recently updated to the latest ChefDK which is 2.5.3.  They bundle a newer version of the kitchen-ec2 gem which appears to break backward compatible functionality.  The issue at hand is that I work in an environment where the AWS SSH managed keys are disallowed.  In fact, my colleague at the time submitted a patch against kitchen-ec2 to not require this attribute be set: https://github.com/test-kitchen/kitchen-ec2/pull/269.  

Would it be possible to have a flag that disabled the creation of the key automagically to restore the previous functionality .... ?

cheeseplus:
Fixes #391. When adding the new auto-generation feature of keys/security groups, the option to _not_ pass any SSH key id at all was lost. This means that `aws_ssh_key_id` now has three states:

- if `nil` we auto-create a key-pair
- if `"/path/to/key.pem"` or `ENV["AWS_SSH_KEY_ID"]` is set, we use specified key
- if `"_disable"`, don't provide any key-pair at all - this requires the remote image/instance to have this configured out of band (baking, user_data)

### TODO

- [x] Update the docs once the triple-option is settled upon.


spec/kitchen/driver/ec2_spec.rb
```diff
@@ -678,6 +678,16 @@
       include_examples "common create"
     end
 
+    context "with key pair configured to false" do
+      before do
+        config[:aws_ssh_key_id] = false
+        expect(driver).to receive(:submit_server).and_return(server)
```
robbkidd: Maybe include a negative expectation for setting `aws_ssh_key_id` to `false`?

```
expect(driver).to_not receive(:create_key)
```

Looking at the spec, I found myself wondering what was different.

cheeseplus: The key won't get created if you provide it either

pantocrator27:
Just throwing this out there .... should `ENV["AWS_SSH_KEY_ID"]` also be allowed to be set to `false`.

cheeseplus:
No, env vars have no concept of boolean - they are always strings and we're going to respect whatever a person sets there and fail if it's bad data.

cheeseplus:
@pantocrator27 so we've figured out the fix (mostly) but bike shedding on the option, would you prefer:

``
aws_ssh_key_id: false
```
OR
```
aws_ssh_key_id: disable #perhaps a symbol even, :disable
```

**Comment by robbkidd** [comment]
@pantocrator27 Updated this PR with what `:disable` might look like.

**Comment by cheeseplus** [comment]
That was pretty much our thinking too but just wanted to shop it out since, as of now, you are the primary use case.

File: spec/kitchen/driver/ec2_spec.rb [hunk 2]
```diff
@@ -658,26 +683,52 @@
...
-        fake_fd = double()
-        fake_file = double()
-        allow(File).to receive(:sysopen).and_call_original
-        expect(File).to receive(:sysopen).with("/kitchen/.kitchen/instance_name.pem", kind_of(Numeric), kind_of(Numeric)).and_return(fake_fd)
-        allow(File).to receive(:open).and_call_original
-        expect(File).to receive(:open).with(fake_fd).and_yield(fake_file)
-        expect(fake_file).to receive(:write).with("RSA PRIVATE KEY")
+      context "with no AWS-managed ssh key pair configured, creates a key pair to use" do
+        before do
+          config.delete(:aws_ssh_key_id)
```
robbkidd: 🤔 This should probably be `config[:aws_ssh_key_id] = nil` to accurately reflect the default config where the key is present and its value is `nil`. `config.delete(:aws_ssh_key_id)` as setup is technically testing when the key is not present.

coderanger:
Hmm, I am :-/ on using disable since we shouldn't do stuff that requires symbols (as opposed to strings) and magic string values are usually a thing that will bite people later.

robbkidd:
@coderanger Do you have a suggestion for a better way to disable the current `SET KEY NAME || GENERATE KEY` behavior? @cheeseplus and I talked about Yet Another Config Option and didn't like the taste of that either.

cheeseplus:
I'm not super opposed to `false` as it fits in with everything else in the ecosystem even if semantically it's not great.

pantocrator27:
At this point ... I am not going to look a gift horse in the mouth :) or ask you guys to do something contrary to the set of current coding standards

pantocrator27:
As the customer, I will be equally pleased with either approach so long as it is documented.  Please and thanks!

robbkidd:
Alternate proposal: `config[:aws_ssh_key_id] = ⛔️` :trollface:

pantocrator27:
@robbkidd can that be done :)

pantocrator27:
also I hate to ask this, could this make its way into the next release of ChefDK ... I don't know how that process would work?

cheeseplus:
The deps in ChefDK just need to get bumped - the catch being that the ChefDK release is on it's own monthly schedule so if you wanted it sooner you'd have to consume a build from the `current` channel once the Gemfile.lock has been bumped.

pantocrator27:
thanks for the info @cheeseplus!

coderanger:
I think I've convinced myself that a magic string is fine but what if we do `_disabled` instead just to make it clear it's a special value? And re: `to_sym`, better to do `.to_s == 'whatever'` since `to_sym` involves permanent allocs ;-)

robbkidd:
Updated for `"_disabled"`.

coderanger:
Just needs some text in the readme to document it :) I can do that post merge if you're low on cycles since you already put in a :100: effort <3

robbkidd:
Regarding ...

> should `ENV["AWS_SSH_KEY_ID"]` also be allowed to be set to [`_disabled`]?

As things currently stand in this PR, if the environment variable is set to `"_disabled"`, that value will carry in and apply in the same way as if you had set `aws_ssh_key_id: _disabled` in kitchen.yml.

robbkidd:
@coderanger I'm working on the README update right now! Thanks for the offer, though.

pantocrator27:
I personally feel it makes sense

cheeseplus:
We've got two thumbs and since I started the PR I can't approve - :shipit:

PR #400, opened 2018-05-18
Fix dynamic key creation

bdwyertech:
Was seeing a weird issue where Ruby was trying to write to the file before it had actually been created on the file system.... Lost the error output, but it was something along the lines of `File not open for writing`

Moving this all into one code block has resolved the issue.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -785,10 +785,9 @@ module Kitchen
         resp = ec2.client.create_key_pair(key_name: "kitchen-#{name_parts.join('-')}")
         state[:auto_key_id] = resp.key_name
         info("Created automatic key pair #{state[:auto_key_id]}")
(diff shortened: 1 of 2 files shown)


coderanger:
Hmm, now I'm trying to remember why I did that.

bdwyertech:
That's the exact error I was seeing @cheeseplus ... Idk why the tests see this as not creating the file... File.open is just File.new which accepts a block. Guess I'll dick around with this on my Mac next week.

bdwyertech:
@coderanger maybe because Ruby was weird at the time regarding file permissions and `File.open` vs `File.new`?  Who knows.  There was a little comment in-line above about syscall and file permissions, but this seems to achieve the same, tested on the Mac too.

jjasghar:
I just ran into an issue with this, this fixed it. Can we get this released in a new gem?

cheeseplus:
https://github.com/test-kitchen/kitchen-ec2/pull/402

Issue #360, opened 2017-12-09
Atomic Tags


freimer:
kitchen-ec2 still does not use atomic tags (tags applied at instance creation), and instead tags the instance, and volume, after the instance is created.  This prevents IAM policy from requiring certain tags at instance creation, restricting the tags to specific tags, or values of those tags.

Request is to change the instance creation to include the tags in the RunInstances request, per:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html
See TagSpecification.N in Request Parameters.

TagSpecification is just a list of Tag objects, which are just Key and Value per the normal AWS tag mechanism.

Unfortunately, atomic tagging is not available in RequestSpotInstances.  The solution should automatically apply the tags for on-demand instances, and only apply the tags after the instance is created if spot instances are requested.

polliard:
+1

cheeseplus:
We've requested that you update rebase the existing PR https://github.com/test-kitchen/kitchen-ec2/pull/307 and we can get that added.

mikewertheimer:
+1

PR #414, opened 2018-09-06
Implementation of Atomic tagging

sumitag:
Potential implementation using a flag for #360

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: bundler
 sudo: false
 branches:
   only:
-  - master
+  - atomic-tags
 rvm:
   - 2.3.7
   - 2.4.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: bundler
 sudo: false
 branches:
   only:
-  - atomic-tags
+  - master
 rvm:
   - 2.3.7
   - 2.4.4
--- a/kitchen-ec2.gemspec
+++ b/kitchen-ec2.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "test-kitchen", "~> 1.4", ">= 1.4.1"
   gem.add_dependency "excon"
   gem.add_dependency "multi_json"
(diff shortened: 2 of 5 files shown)


coderanger:
There's not really any reason to introduce a new option. It should use in-stream tags for non-spot instance and dumb-slow-tags for spot instances.

(also almost everyone should probably be using spot instances for kitchen-ec2 which is part of why I've not really bothered fixing this)

.travis.yml
```diff
@@ -3,7 +3,7 @@ cache: bundler
 sudo: false
 branches:
   only:
-  - master
+  - atomic-tags
```
coderanger: This doesn't work, also never do this in a FOSS project.

kitchen-ec2.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "test-kitchen", "~> 1.4", ">= 1.4.1"
   gem.add_dependency "excon"
   gem.add_dependency "multi_json"
-  gem.add_dependency "aws-sdk", "~> 2"
+  gem.add_dependency "aws-sdk", "~> 3"
```
coderanger: This needs to be x-ref'd with ChefDK.

.travis.yml
sumitag: Absolutely, I was just running through travis for my local branch and forgot to revert it.

kitchen-ec2.gemspec
sumitag: ChefDK is using ~> 2, I am going to change it back.

sumitag:
At our work we are using normal instances and not spot instances and also want to restrict instance creations with valid tags using IAM.

I will implement in-stream tags for non spot instances.

sumitag:
@coderanger I have updated the branch, can you please have a look?

clintoncwolfe:
I'm looking to continue this work; @sumitag do you mind if I adopt this PR?

sumitag:
Please feel free to use the PR

tas50:
This has been resolved with https://github.com/test-kitchen/kitchen-ec2/pull/496

PR #427, opened 2018-10-26
State file add ips mb

mbaitelman:
Add private and public IP's to the state file

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -261,8 +261,8 @@ module Kitchen
           image_data = ec2.client.describe_images({ :image_ids => [server.image_id] })[0][0]
 
           state[:server_id] = server.id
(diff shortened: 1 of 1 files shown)


cheeseplus:
Why do we need this? We only need one IP to communicate to the instance and never is kitchen switching between the two. This is also breaking several tests so even if this was a feature we needed it can't be merged without fixing the tests.

mbaitelman:
I found it useful to have the IP esp when debugging network issues in AWS.
I was going to spend some time to try to understand why those tests are failing.
If this is something that there isnt a community benefit for I will close the request.

mbaitelman:
Closing the pull request as the use case is quite small.

mbaitelman:
Learned some more things over the year. 
Turns out the [interface ](https://github.com/test-kitchen/kitchen-ec2#interface) value is what I needed.

Issue #431, opened 2018-11-27
Kitchen is using ARM architecture for Ubuntu by default


axl89:
Since today, and without changes on my kitchen (`v1.19.2`) configuration file, kitchen spawned ARM machines and failed.

## Relevant section of the configuration file

```
---
driver:
  name: ec2
  region: eu-west-1
  instance_type: t2.medium
  associate_public_ip: true
  interface: dns
  aws_ssh_key_id: AWSIreland
  block_duration_minutes: 60
  block_device_mappings:
    - device_name: /dev/sda1
      ebs:
        volume_size: 12
        delete_on_termination: true


transport:
  ssh_key: ~/.ssh/AWSIreland.pem
  connection_timeout: 10
  connection_retries: 5
  username: ubuntu

provisioner:
  name: chef_zero
  product_name: chef
  client_rb:
    environment: integration

verifier:
  name: inspec

platforms:
  - name: ubuntu-16.04
  - name: ubuntu-18.04
```

## CI output

```
$ kitchen test ^hextron-13.* -d always
-----> Starting Kitchen (v1.19.2)
-----> Cleaning up any prior instances of <hextron-13-full-test-ubuntu-1604>
-----> Destroying <hextron-13-full-test-ubuntu-1604>...
       Finished destroying <hextron-13-full-test-ubuntu-1604> (0m0.00s).
-----> Testing <hextron-13-full-test-ubuntu-1604>
-----> Creating <hextron-13-full-test-ubuntu-1604>...
       Detected platform: ubuntu version 16.04 on arm64. Instance Type: t2.medium. Default username: ubuntu (default).
       If you are not using an account that qualifies under the AWS
free-tier, you may be charged to run these suites. The charge
should be minimal, but neither Test Kitchen nor its maintainers
are responsible for your incurred costs.

-----> Destroying <hextron-13-full-test-ubuntu-1604>...
       Finished destroying <hextron-13-full-test-ubuntu-1604> (0m0.00s).
-----> Cleaning up any prior instances of <hextron-13-full-test-ubuntu-1804>
-----> Destroying <hextron-13-full-test-ubuntu-1804>...
       Finished destroying <hextron-13-full-test-ubuntu-1804> (0m0.00s).
-----> Testing <hextron-13-full-test-ubuntu-1804>
-----> Creating <hextron-13-full-test-ubuntu-1804>...
[...]

PR #433, opened 2018-11-27
Adding support for arm64 architecture

tyler-ball:
This gives us basic support for `arm64` instances. Users can specify the architecture or a TK platform name like `ubuntu-18.04-arm64` to create these. They do need to specify the instance type as `a1.*` for AWS to successfully create it.

It also gets rid of some unsupported EC2 architectures

Potential improvements:
[ ] Set the default instance type to `a1.medium` for arm64 architecture
[ ] Better error message if the user tries to specify arm64 architecture and does not specify a `a1.*` instance type
[ ] Tests around this


robbkidd:
> - Tests around this

Yes please!

> - Set the default instance type to a1.medium for arm64 architecture
> - Better error message if the user tries to specify arm64 architecture and does not specify a a1.* instance type

On the one hand, these would seem to increase user-friendliness. On the other, we'll be chasing AWS EC2 image and instance type permutations.

cheeseplus:
We've always been chasing those and by the nature of the driver itself we'll always be chasing so not much different from the status quo.

Gemfile
```diff
@@ -3,9 +3,6 @@ source "https://rubygems.org"
 # Specify your gem"s dependencies in kitchen-ec2.gemspec
 gemspec
 
-gem "winrm-transport"
-gem "winrm-fs"
```
robbkidd: 👍 I had done this locally while troubleshooting #431 yesterday, but hadn't hunted down whether they were really needed.

spec/kitchen/driver/aws/image_selection_spec.rb
```diff
@@ -32,7 +32,8 @@ def new_instance(platform_name: "blarghle")
       :provisioner => Kitchen::Provisioner::Dummy.new,
       :transport => Kitchen::Transport::Dummy.new,
       :verifier => Kitchen::Verifier::Dummy.new,
-      :state_file => Kitchen::StateFile.new("/nonexistent", "suite-name-#{platform_name}")
+      :state_file => Kitchen::StateFile.new("/nonexistent", "suite-name-#{platform_name}"),
+      lifecycle_hooks: Kitchen::LifecycleHooks.new({})
```
robbkidd: Can we make these consistent `:key =>` or `key:` after we verify everything passes?

spec/kitchen/driver/aws/standard_platform_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      virtualization_type: "hvm",
+      name: "ubuntu"
+    ) }
+    let(:images) { [img1, img2, img3, img4] }
+    let(:sorted_images) { [img4, img3, img2, img1] }
+    let(:standard_platform) { klass.new(nil, nil, nil, nil) }
+
+    it "correctly sorts the images" do
+      expect(standard_platform.send(:sort_images, images)).to eq(sorted_images)
+    end
```
robbkidd: sweet

✨ It's a start! ✨

Gemfile
tyler-ball: Yeah - I didn't see anywhere they were used, and `winrm-transport` has been completely deprecated

robbkidd: buh-bye deprecated code

Issue #417, opened 2018-09-25
Spot instance use case


vmiszczak-teads:
Hello,

We are using spot instances with `kitchen-ec2` and it does not fit ours needs well at the moment.
We often have timeouts getting the instance_type/subnet/price we ask for.

Having discussed with AWS guys, it looks like working with spots is more a workflow like :
- have a pool of instance_type, not just 1
- have a pool of subnets, not just 1
- bid at on-demand price

It seems not possible to do this in the current state.

We see at least 2 approaches to the issue:
- externally try one set of value, if fail, retry with another set
It ideally requires a distinct exit code to distinguish between failing getting a spot and a failed Chef test.
 
- modify kitchen-ec2 to support this kind of spot workflow
It could be done compatible with current state by adding options or incompatible by changing some current options behavior/meaning.

What you guys think about this use case and solutions?

vmiszczak-teads:
I may be wrong but the only way to request a spot is to provide an instance type, a subnet and a spot price. I also use some other settings like block_duration_minutes, region, subnet_id.
Because the combination of settings is fixed, I sometimes get a spot, sometimes not, depending of availability on AWS.
Is there another way to request spots ?
I didn't try other settings as I didn't see other options in the manual.

I'd like to provide an array of instance type, an array of subnets, no spot price (meaning biding on-demand price), a priority list(I want to try another subnet before moving trying another instance type, having the choice is nice), a timeout per setting combination and let the driver try everything until it can get a spot or fail.

cheeseplus:
We can look at the existing implementation but as described thats a lot of gold plating for a fairly niche feature. We'll happily review a PR for something along these lines but that's a ton of configurables and not something we're likely to pursue in the near future.

vmiszczak-teads:
I'll probably work on a PR.
Do you mind breaking compatibility (using versioning)? 
Or a new system should be backward compatible with the current implementation?

cheeseplus:
We adhere to SemVer so breaking changes would necessitate a major release. I'd prefer that we don't arbitrarily bump to 3.x just for this one feature but it depends on the scope.

vmiszczak-teads:
I made a first attempt to do this:
https://github.com/teads/kitchen-ec2/pull/1

`instance_type` and `subnet_id` can be arrays and I added `spot_wait` timeout to not mess with other timeouts.
[...]

PR #437, opened 2018-12-17
Allow asking for multiple instance types and subnets

vmiszczak-teads:
This is a solution for https://github.com/test-kitchen/kitchen-ec2/issues/417.
`instance_type` and `subnet` can be arrays.
`spot_price` can be 'on-demand'.

It allows trying harder getting a spot before failing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ region.
 The EC2 [instance type][instance_docs] (also known as size) to use.
 
 The default is `t2.micro` or `t1.micro`, depending on whether the image is `hvm`
-or `paravirtual`. (`paravirtual` images are incompatible with `t2.micro`.)
+or `paravirtual`. (`paravirtual` images are incompatible with `t2.micro`). When working with spots, you can provide an array of instance types in which case the driver will try each type until it can get a spot.
 
 #### `security_group_ids`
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -271,32 +284,31 @@ The EC2 IAM profile name to use. The default is `nil`.
...
 }
 ```
 
 See [AWS documentation](https://aws.amazon.com/de/blogs/security/granting-permission-to-launch-ec2-instances-with-iam-roles-passrole-permission/) for more details.
 
-
 #### `spot_price`
 
-The price you bid in order to submit a spot request. An additional step will be required during the spot request process submission. If no price is set, it will use an on-demand instance.
+The price you bid in order to submit a spot request. An additional step will be required during the spot request process submission. If no price is set, it will use an on-demand instance.
```
tas50: This should stay on a single line so we can let soft wrapping utilize as much screen real estate as is available

README.md
```diff
@@ -452,14 +464,14 @@ platforms:
   - name: windows-2012r2
   - name: windows-2016
 
-suites:
+? suites
```
tas50: This seems wrong

vmiszczak-teads:
@tas50 thank you for the review. I disabled my IDE "format on save" feature so the original linting is preserved.

dpacaud:
Hi @tas50 : any way to see this merged ?
thanks !

vmiszczak-teads:
Hi,
What can I do to help merge/improve this?

cheeseplus:
Not sure when the next release will go out but this looks good and don't see any reason why we shouldn't take this one.

dpacaud:
thanks !

PR #446, opened 2019-02-22
Make ingress ip range configurable

dpattmann:
It would be nice the change the CIDR block to a special IP address when the SG is created automatically.


cheeseplus:
I don't see any reason (yet) that we'd not merge this as it looks well thought out and has tests. 👍

tas50:
Thanks @dpattmann

PR #454, opened 2019-05-01
Fix aws-sdk-ec2 deprecation warnings

Annih:
The `[]` operator on Instance::Collection is deprecated.
Yielding the instance waiter to a block is deprecated.


tas50:
Thanks @Annih

PR #458, opened 2019-06-12
Security Group Filter Enhancement

bdwyertech:
Adds support for searching for multiple groups, as well as searching by group name.  Sometimes SG's get created without tags, but they will always have a name.

Also, in a large enterprise environment you may need to leverage multiple security groups.  Using a list of search patterns is more readable and potentially portable across the account boundary.

This lets you do this:
```yaml
  security_group_filter:
    - name: 'AWS-Egress'
    - tag: 'Name'
      value: 'MyApplicationSG'
    - tag: 'Name'
      value: 'MyApplicationDatabaseSG'
    - tag: 'Name'
      value: 'MyOtherSG'
```
Instead of this
```yaml
security_group_ids:
  - sg-abc12345657891011
  - sg-accxsdsl2130dkj213
  - sg-0aabc123412321232
```

The code is backwards compatible with pre-existing `security_group_filter` functionality.

### DIFF
--- a/lib/kitchen/driver/aws/instance_generator.rb
+++ b/lib/kitchen/driver/aws/instance_generator.rb
@@ -89,8 +89,7 @@ module Kitchen
               if security_group
                 security_groups.push(security_group.group_id)
               else
-                raise "The group tagged '#{sg_filter[:tag]} " +
-                  "#{sg_filter[:value]}' does not exist!"
+                raise "A Security Group matching the following filter could not be found:\n#{sg_filter}"
               end
             end
             config[:security_group_ids] = security_groups
--- a/lib/kitchen/driver/aws/instance_generator.rb
+++ b/lib/kitchen/driver/aws/instance_generator.rb
@@ -62,22 +62,38 @@ module Kitchen
           end
 
           if config[:security_group_ids].nil? && config[:security_group_filter]
-            security_group = ::Aws::EC2::Client
-                .new(region: config[:region]).describe_security_groups(
-                filters: [
-                    {
-                        name: "tag:#{config[:security_group_filter][:tag]}",
-                        values: [config[:security_group_filter][:value]],
-                    },
+            security_groups = []
+            filters = [config[:security_group_filter]].flatten
+            filters.each do |sg_filter|
+              r = {}
+              if sg_filter[:name]
+                r[:filters] = [
+                  {
+                    name: "group-name",
+                    values: [sg_filter[:name]],
+                  },
                 ]
-            )[0][0]
+              end
 
-            if security_group
-              config[:security_group_ids] = [security_group.group_id]
-            else
-              raise "The group tagged '#{config[:security_group_filter][:tag]} " +
-                "#{config[:security_group_filter][:value]}' does not exist!"
+              if sg_filter[:tag]
+                r[:filters] = [
+                  {
+                    name: "tag:#{sg_filter[:tag]}",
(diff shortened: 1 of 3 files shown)


lamont-granquist:
I don't have enough domain knowledge to review the shape of this API, code looks fine though, i was gonna suggest `each_with_object` might be used, but it looks like there's more complicated data wrangling going on inside the loop.

lib/kitchen/driver/aws/instance_generator.rb
```diff
@@ -62,22 +62,38 @@ def ec2_instance_data # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
...
-                        values: [config[:security_group_filter][:value]],
-                    },
+            security_groups = []
+            filters = [config[:security_group_filter]].flatten
+            filters.each do |sg_filter|
+              r = {}
+              if sg_filter[:name]
+                r[:filters] = [
+                  {
+                    name: "group-name",
```
afiune: What would happen if we have multiple `name` entries? 🤔 Something like:
```
security_group_filter:
  - name: 'AWS-A'
  - name: 'AWS-B'
  - tag: 'Name'
    value: 'my-app'
```
Is it allowed to have multiple `group-name` entries or should we restrict/check to have only one inside `filters`?

bdwyertech: @afiune multiple group name entries are fine.  In the case there it would grab all 3 security groups.

bdwyertech: I think the better answer to your question is that it's one search per array entry.  So yeah youd be searching 3 times and being dropped into all 3 security groups.  This is the intended behavior.

Would love to see this merged into ChefDK/Workstation prior to the next release-- I've already poked my Chef rep about it in June.

lib/kitchen/driver/aws/instance_generator.rb
```diff
@@ -62,22 +62,38 @@ def ec2_instance_data # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
...
+                ]
+              end
+
+              security_group = ::Aws::EC2::Client.new(region: config[:region]).describe_security_groups(r)[0][0]
+
+              if security_group
+                security_groups.push(security_group.group_id)
+              else
+                raise "The group tagged '#{sg_filter[:tag]} " +
+                  "#{sg_filter[:value]}' does not exist!"
```
afiune: I think here we are assuming that `tag` and `value` are always provided but now we accept `name` so I think we need to update this error to know when we are looking for a group-name or a tag.

lib/kitchen/driver/aws/instance_generator.rb
afiune: Makes perfect sense! Thanks for that answer. We are looking into merging this PR 💯

lib/kitchen/driver/aws/instance_generator.rb
bdwyertech: Agreed.  We could probably just spit out the entire object instead.  Will update the PR tmm.

bdwyertech: Code has been updated, error output now looks like
```
>>>>>> ------Exception-------
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #create action: [A Security Group matching the following filter could not be found!{:tag=>"Name", :value=>"MyCoolGroup"}] on default-centos7
```

PR #482, opened 2020-02-14
Don't crash upon destroy if instance is already dead

kamaradclimber:
On some environment, instances are automatically destroyed after a while
(to avoid leaking test instance that would be billed for a long time).

This patch, ignore errors from ec2 when instance id is not found upon
destroy.

It will solve the long running issue where user runs `kitchen test` >1d
after running a `kitchen test --destroy=never` for instance

This is a cherry-pick of https://github.com/criteo-forks/kitchen-ec2/pull/8/commits to latest kitchen-ec2 version

Change-Id: I331c01b43c5797426361d0268f3feefc88d1e8ce


tas50:
Nice fix. Thanks @kamaradclimber

PR #484, opened 2020-03-18
Remove obsolete setting

arothian:
Cleaning up the example to remove `require_chef_omnibus` that was removed in https://github.com/test-kitchen/kitchen-ec2/commit/a5273014b9b30dd6c6144c6a555085fe8245883d


Xorima:
Thanks for that @arothian

PR #485, opened 2020-03-25
Use multiple results from subnet_filter for spot requests

benclark:
We were having issues finding spot capacity for our test instances because the launch specification was only using the first result from `subnet_filter` searches.

I slightly reworked the `submit_spots` method to replace `config[:subnet_id]` with the results of the filter prior to expanding the configs.

I left the behavior within the `ec2_instance_data` method the same, i.e. it will continue to return the first result of the filter search.

### DIFF
--- a/lib/kitchen/driver/aws/instance_generator.rb
+++ b/lib/kitchen/driver/aws/instance_generator.rb
@@ -39,10 +39,8 @@ module Kitchen
           @logger = logger
         end
 
-        # Transform the provided config into the hash to send to AWS.  Some fields
-        # can be passed in null, others need to be ommitted if they are null
-        def ec2_instance_data # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
-          # Support for looking up security group id and subnet id using tags.
+        # Query AWS for subnet_ids matching subnet_filter config.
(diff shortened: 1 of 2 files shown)


tas50:
@benclark Does the new logic introduced in https://github.com/test-kitchen/kitchen-ec2/pull/489 handle this for you?

benclark:
@tas50 - Not exactly, but it's close. The problem we were running into wasn't that our subnets were too full -- in fact, all our test builds run in dedicated subnets so there are typically no other instances running -- but that (for example) there wasn't any m4.large spot capacity in us-east-1a, and the spot requests were failing. With my patch, it at least tries spot requests in us-east-1b, us-east-1c, etc., before giving up. I'm not certain the logic in #489 will help in our case, since the first subnet it finds would have plenty of available IPs.

I'm happy to take another swing at this PR if my approach wasn't the right one. My first thought was to refactor to use spot fleet requests and let AWS decide which instances go where, but when I looked in the kitchen-ec2 code, that seemed like too big of a lift.

benclark:
Looks like my original issue was fixed by #499, and in a better way, so I'm closing this.

PR #489, opened 2020-04-23
Least Populated Subnet

bdwyertech:
Select the least-populated subnet if we have multiple matches -- previously we just took the first match.

VPC may span multiple regions -- subnets are region specific.  This lets you pick the least populated VPC.

This should help to distribute the test-kitchen load more evenly across multi-az VPC's while maintaining full backward compatibility.

We've noticed intermittent problems internally with our cookbook generator where folks do not change the default (e.g.   org_*_dmz => always resolves to org_us-east-1a-dmz).  Kitchens sometimes fail either because of spot instance unavailability, ip starvation, etc.

@sumitag @chefbob @danielcbright

### DIFF
--- a/lib/kitchen/driver/aws/instance_generator.rb
+++ b/lib/kitchen/driver/aws/instance_generator.rb
@@ -56,8 +56,10 @@ module Kitchen
             ).subnets
             raise "The subnet tagged '#{config[:subnet_filter][:tag]}:#{config[:subnet_filter][:value]}' does not exist!" unless subnets.any?
 
-            vpc_id = subnets[0].vpc_id
-            config[:subnet_id] = subnets[0].subnet_id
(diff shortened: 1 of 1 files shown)


bdwyertech:
The driver also uses similar logic over here to create a security group if one is not supplied:
https://github.com/test-kitchen/kitchen-ec2/blob/6d646b38cae3ad292aa187619ac4d7e81f4841e6/lib/kitchen/driver/ec2.rb#L792

I could not for the life of me figure out how to mock the following in RSpec:
```ruby
subnet = subnets.sort_by { |s| s[:available_ip_address_count] }.last
```

chefbob:
Nice.  Thanks for putting this together.  Definitely would come in handy.

tas50:
Nice little addition there. Thanks @bdwyertech

PR #364, opened 2018-01-22
WIP: Tag server / volumes during launch

freimer:
Needs testing.  Created a new branch as there were too many conflicts.

PR #496, opened 2020-06-17
Tag on-demand instances at creation time

### LINKED ISSUE #360: Atomic Tags

### LINKED ISSUE #464: aws ec2 and EBS volume tags at instance creation
Tags should be applied at the time instance creation without this instance creation is denied by strict tags policy.

### LINKED ISSUE #480: Instances created without specific tags are terminated by AWS SCP.
We have an issue where a new AWS SCP requires certain tags on instances cannot be created.  It appears to me that kitchen-ec2 is creating the instance without tags and then trying to tag it.
- Is this correct?
- Is it implemented this way for a specific reason?
- Anyone else have similar issue and possibly a work around?

Thanks

clintoncwolfe:
# Description

This is started as an adoption of #414 by @sumitag, adding tests and rebasing. #414 applies tags at instance creation time to on-demand instances. 

This PR then goes on to convert spot instance requests to use the `create_instances()` API call, which allows tagging at creation time. This allows all tagging to happen at creation time, so all post-hoc tagging code is removed.

There are some fallout effects from using `create_instances()` instead of `spot_instance_request()`:
 * Since it is not a spot instance request, the is no Spot Instance Request ID to report to the user.
 * Spot Instance Requests internally retry the request until it succeeds or the request expires. `create_instances()` is one-shot, so kitchen-ec2 must explicitly retry the call.
 * ~There is a feature that allows the user to set the spot price to "on-demand" to get the on-demand price; internally this sets the bid price to an empty string. This is not allowed in the `create_instances()` API. **Currently this is not solved, and the feature is broken pending discussion.**~ Update: When on-demand is specified, the max_price field may simply be omitted, causing the bid price to be set to the on-demand price.

## Issues Resolved

Fixes #360 
Fixes #480 
Fixes #464

I suggest PRs #364 and #414 be closed.

## Check List

- [x] All tests pass. See TESTING.md for details.
- [x] New functionality includes testing.
- [x] New functionality has been documented in the README if applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,6 @@ The Hash of EC tag name/value pairs which will be applied to the instance.
 
 The default is `{ "created-by" => "test-kitchen" }`.
 
-Tags are applied post creation for spot instances and during creation for standard instances.
 
 #### `user_data`
 
--- a/README.md
+++ b/README.md
@@ -276,6 +276,8 @@ The Hash of EC tag name/value pairs which will be applied to the instance.
 
 The default is `{ "created-by" => "test-kitchen" }`.
 
+Tags are applied post creation for spot instances and during creation for standard instances.
+
 #### `user_data`
 
 The user_data script or the path to a script to feed the instance.
--- a/lib/kitchen/driver/aws/instance_generator.rb
+++ b/lib/kitchen/driver/aws/instance_generator.rb
@@ -115,7 +115,21 @@ module Kitchen
             key_name: config[:aws_ssh_key_id],
             subnet_id: config[:subnet_id],
             private_ip_address: config[:private_ip_address],
+            min_count: 1,
+            max_count: 1,
           }
+
+          if config[:tags] && !config[:tags].empty?
+            tags = config[:tags].map do |k, v|
+              # we convert the value to a string because
+              # nils should be passed as an empty String
+              # and Integers need to be represented as Strings
+              { key: k, value: v.to_s }
+            end
+            instance_tag_spec = { resource_type: "instance", tags: tags }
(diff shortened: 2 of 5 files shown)


clintoncwolfe:
After some conversation with @tyler-ball, we think we may be able to use the AWS run_instance() call to tag spot instances at creation time as well. I'm placing this into draft status while I do a rewrite.

README.md
```diff
@@ -276,6 +276,8 @@ The Hash of EC tag name/value pairs which will be applied to the instance.
 
 The default is `{ "created-by" => "test-kitchen" }`.
 
+Tags are applied post creation for spot instances and during creation for standard instances.
```
tyler-ball, line 279:
```suggestion
```

We can just get rid of this note now since all tags are added during creation, correct?

tas50:
Way cool. Thanks @clintoncwolfe

PR #499, opened 2020-06-30
Spot Instances - Cascading Subnet Filter Support

bdwyertech:
## Description

Spot instances currently support cascading through an explicitly configured list of subnets.  This adds support for using a subnet filter to match multiple subnets, and behave in the same way.  Hardcoded resource ID's are often less readable and less portable than a filter.

## Related

#417 
#437 

## Check List

- [ ] All tests pass. See TESTING.md for details.
- [ ] New functionality includes testing.
- [ ] New functionality has been documented in the README if applicable.

### DIFF
--- a/lib/kitchen/driver/ec2.rb
+++ b/lib/kitchen/driver/ec2.rb
@@ -424,7 +424,31 @@ module Kitchen
       def submit_spots
         configs = [config]
         expanded = []
-        keys = %i{instance_type subnet_id}
+        keys = %i{instance_type}
+
+        unless config[:subnet_filter]
+          # => Use explicitly specified subnets
+          keys << :subnet_id
+        else
+          # => Enable cascading through matching subnets
+          client = ::Aws::EC2::Client.new(region: config[:region])
+          subnets = client.describe_subnets(
+            filters: [
+              {
+                name: "tag:#{config[:subnet_filter][:tag]}",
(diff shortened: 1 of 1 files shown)


tas50:
@bdwyertech Can you rebase this against master please?

bdwyertech:
Sure. Give me a sec.

bdwyertech:
Alright, should be rebased against master now @tas50  -- I think Ruby 2.4 was failing before due to usage of Array.append -- what was the syntax for that in 2.4?  I don't even remember.

https://github.com/test-kitchen/kitchen-ec2/pull/499/files#diff-c5498dad76244656522b77033980c6a6R431

lib/kitchen/driver/ec2.rb
```diff
@@ -424,7 +424,31 @@ def expand_config(conf, key)
...
         configs = [config]
         expanded = []
-        keys = %i{instance_type subnet_id}
+        keys = %i{instance_type}
+
+        unless config[:subnet_filter]
+          # => Use explicitly specified subnets
+          keys << :subnet_id
+        else
+          # => Enable cascading through matching subnets
```
tas50, line 433: it seems like this whole part could be extracted into a method and mocked / tested to some degree. I know we have massive methods already in this driver, but I'd like to try to prevent the sprawl going forward and slowly tackle the sprawl we've already shipped at a later date.

PR #509, opened 2020-08-19
Allow multiple ip addresses to be specified when creating a security group

trainsushi:
# Description

Allow `security_group_cidr_ip` option to accept either a string or an array. By providing an array, the user can specify multiple IP addresses for the security group. The default value stays the same `0.0.0.0/0`.

Currently, `security_group_cidr_ip` option only allows one ip address to be added to the temporary security group. Oftentimes, allowing more than one ip address is necessary because load balancers in cloud have multiple ips that change frequently. For instance, when test-kitchen runs on TravisCI, the temporary security group created by kitchen-ec2 should include all the public ips of Travis since kitchen on Travis should be able to connect to a test instance on your AWS VPC. Using an existing security group with hard-coded public ips is not ideal because their public ips are dynamic and can change.
TravisCI public ip info: https://docs.travis-ci.com/user/ip-addresses/

## Issues Resolved

n/a

## Check List

- [x] All tests pass. See TESTING.md for details.
- [x] New functionality includes testing.
- [x] New functionality has been documented in the README if applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,11 +237,11 @@ security_group_filter:
     value: 'MyOtherSG'
 ```
 
-#### `security_group_cidr_ips`
+#### `security_group_cidr_ip`
 
-An Array of EC2 [security group][group_docs] ip(s), in CIDR block format, to use when creating the security group.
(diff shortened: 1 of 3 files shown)


trainsushi:
@clintoncwolfe Thank you for your feedback. As per your suggestion, I changed `security_group_cidr_ip` to accept either an array or a string. The default value stays the same `0.0.0.0/0`, and this change should be backward compatible.

Issue #479, opened 2020-01-22
Kitchentest instance with IMDSv2 


raghavvidya:
Hello, Is there any option for running the Kitchen test with AWS metadata version 2(IMDVs2) ?

See the announcement https://aws.amazon.com/about-aws/whats-new/2019/11/announcing-updates-amazon-ec2-instance-metadata-service/ and the user guide https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

tas50:
@raghavvidya Can you provide more information on what you're hoping to do with Test Kitchen and the AWS metadata version 2 and what you can't do at this point?

TheSAS:
The question is how to pass this configuration into the kitchen:
```
  metadata_options {
    http_endpoint               = "enabled"
    http_tokens                 = "required"
    http_put_response_hop_limit = 1
  }
```
https://docs.aws.amazon.com/cli/latest/reference/ec2/run-instances.html (--metadata-options)

fletchowns:
I am wondering this as well - organization has a requirement to enforce IMDSv2 on all EC2 instances but I'm not seeing a way to do that with instances generated by test-kitchen. Did anybody figure this out yet?

chenwany:
Hi, our team is also using kitchen test. We detect that the instances launched have IMDSv1 calls, is there any timeline for kitchen to transition to AWS metadata version 2(IMDVs2)? It may potentially block customers who disable IMDSv1 for security consideration.

PR #571, opened 2021-12-23
Optionally enforce IMDSv2 on kitchen instances.

bobchaos:
# Description

Provides an option to enforce IMDSv2 on kitchen instances.

I considered exposing metadata_options directly but this brings it in line with the way it's done in the AWS console.

## Issues Resolved

https://github.com/test-kitchen/kitchen-ec2/issues/479

## Check List

- [X] All tests pass. See TESTING.md for details. 
- [X] New functionality includes testing.
- [X] New functionality has been documented in the README if applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,10 @@ driver:
 
 elastic_network_interface_id's have a format like eni-0545666738adeed14. You can only attach the network interface to instances in the same Availability Zone.
 
+#### `enforce_imdsv2`
+
+Setting this boolean to `true` will enforce IMDSv2 on kitchen instances. See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html for more information on IMDS. Defaults to `false`
+
 ### Disk Configuration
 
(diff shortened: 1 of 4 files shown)


bdwyertech:
Blah, I just added support for similar -- I think we should expose metadata options as well, especially now that control of instance tags via metadata is also exposed via this config block.

https://aws.amazon.com/about-aws/whats-new/2022/01/instance-tags-amazon-ec2-instance-metadata-service/

tas50:
@bobchaos can you rebase this and add the docs to https://github.com/test-kitchen/test-kitchen/blob/main/docs/content/docs/drivers/aws.md

bdwyertech:
> @bobchaos can you rebase this and add the docs to https://github.com/test-kitchen/test-kitchen/blob/main/docs/content/docs/drivers/aws.md

@tas50 the metadata addition would encompass this as well so you can prob close this out.

We need the whole block exposed because it is used to enable instance tags via metadata as well as the IMDSv2 hardening config.

Issue #564, opened 2021-11-03
Support Rocky Linux


tas50:
There are official Rocky Linux AMIs. We should find them and support the same sort of logic that we have for RHEL/CentOS

krayjay:
Has Rocky OS support been implemented yet? Particularly Rocky 8

tas50:
Shipped in 3.17

PR #602, opened 2023-03-21
Adds support for Rocky and AlmaLinux

### LINKED ISSUE #563: Support Alma Linux
There are official Alma Linux AMIs. We should find them and support the same sort of logic that we have for RHEL/CentOS

jakauppila:
# Description

Adds support for Rocky and AlmaLinux

## Issues Resolved

#563 #564 

## Check List

- [x] New functionality includes tests
- [x] All tests pass
- [x] Commit message includes a [Conventional Commit Message](https://www.conventionalcommits.org/en/v1.0.0)


jakauppila:
I wasn't sure on the platform name itself, so I added both `alma`/`almalinux` and `rocky`/`rockylinux`. If there's a preference one way or the other, let me know.

lib/kitchen/driver/aws/standard_platform/rocky.rb
```diff
@@ -0,0 +1,52 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative "../standard_platform"
+
+module Kitchen
+  module Driver
+    class Aws
+      class StandardPlatform
+        # https://docs.aws.amazon.com/linux/al2023/release-notes/relnotes.html
```
RulerOf, line 22: This comment is confusing, I assume you're duplicating the class for AL2023

lib/kitchen/driver/aws/standard_platform/rocky.rb
```diff
@@ -0,0 +1,52 @@
...
+
+          # default username for this platform's ami
+          # @return [String]
+          def username
+            "rocky"
+          end
+
+          def image_search
+            search = {
+              "owner-id" => "792107900819",
```
RulerOf, line 35: I suggest commenting with a link to the docs that establish this as the correct owner ID. I googled the number and the first result isn't what I was looking for.

RulerOf: I'm not a maintainer just figured I'd point out the things I saw since I was reading the code anyway ¯\_(ツ)_/¯

lib/kitchen/driver/aws/standard_platform/rocky.rb
tas50, line 22:
```suggestion
```

Issue #616, opened 2023-11-27
Ruby 3.1 syntax has been introduced


cmparkinson:
## :ghost: Brief Description

PR https://github.com/test-kitchen/kitchen-ec2/pull/610 introduced hash literal value omission, which is a Ruby >= 3.1 syntax change and has broken the gem for all earlier Ruby versions.  The gemspec still specifies Ruby >= 2.7, which no longer holds true.  If this change was intended, it should probably have come with a major version bump to adhere to semver.

## Version

3.17.1

## Steps to Reproduce
Use the gem in a ruby version prior to 3.1.

## Expected Result

The gem should work.

## Actual Result

`SyntaxError` is raised.

cmparkinson:
@damacus Was the `TargetRubyVersion` bump [here](https://github.com/test-kitchen/kitchen-ec2/pull/610/files#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR6) intentional?  The autocorrections have broken the gem for earlier Ruby versions.

damacus:
Thanks for spotting the lack of an upgraded version there. I bumped a few libraries over the past week and obviously forgot this one. I'll do another release shortly 👍🏼

PR #618, opened 2023-11-30
feat: Bump Ruby version to 3.1

damacus:
BREAKING CHANGE: Require Ruby 3.1 or higher to work with this Gem

Signed-off-by: Dan Webb <dan.webb@damacus.io>


damacus:
Fixes #616 where we bumped the Rubocop target version but missed bumping the required Ruby version in the Gem

damacus:
We could target 3.0, So I'm willing to bump this back down to 3.0 from 3.1 if anyone shouts up and disagrees with this

https://endoflife.date/ruby

nuclearsandwich:
@damacus FYI I do think the changes in #610 require 3.1 as I get errors with 3.18 and 3.17.1 using Ruby 3.0.6p216. I've pinned 3.17.0 until I can switch to Ruby 3.1.

PR #623, opened 2024-04-28
feat: add support for IPv6

jjacque:
# Description

Add the possibility to associate an IPv6 with the EC2 instance.
It will obviously require your VPC and subnet to have IPv6 CIDRs defined.

## Issues Resolved

N/A

## Type of Change

Feature addition.

## Check List

- [X] New functionality includes tests
- [ ] All tests pass
- [X] Commit message includes a [Conventional Commit Message](https://www.conventionalcommits.org/en/v1.0.0)


damacus:
We'd want to allow the user to associate IPv6 with a switch (or have it on by default) but allow It to be configured off when it causes trouble.

jjacque:
> We'd want to allow the user to associate IPv6 with a switch (or have it on by default) but allow It to be configured off when it causes trouble. 

Hi, sorry for the late response: you actually want me to reverse the logic ?
Having IPv6 enabled by default and have an option like `disable_ipv6` ?

PR #671, opened 2026-08-23
fix: Stop reporting every create failure as an AMI problem

### LINKED ISSUE #606: Digest initialization failed: initialization error in the specified(Ruby-3.2/OpenSSL-3.0.8 7)
I am trying to run a `kitchen-test` using the latest AWS Codebuild Image https://github.com/aws/aws-codebuild-docker-images/blob/master/al2/aarch64/standard/3.0/Dockerfile 

It is not working with this ruby version but the previous version before ruby 3.2 works fine. I also noticed that this `3.2` was compiled with `openssl v3`

## Environment
* `ruby 3.2.1 (2023-02-08 revision 31819e82c8) [aarch64-linux]`
* `openssl - ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'` **OpenSSL 3.0.8 7 Feb 2023**
* Gemfile
 ```
source 'https://rubygems.org'

gem 'serverspec'
gem 'kitchen-ec2'
gem 'kitchen-inspec'
gem 'test-kitchen'
gem 'rspec-retry'
gem 'inspec-bin'
```

```

## Error
>>>>>> ------Exception-------
agent_1  | >>>>>> Class: Kitchen::ActionFailed
agent_1  | >>>>>> Message: 1 actions failed.
agent_1  | >>>>>>     Failed to complete #create action: [Digest initialization failed: initialization error in the specified region]
agent_1  | >>>>>> ----------------------
agent_1  | >>>>>> Please see .kitchen/logs/kitchen.log for more details
agent_1  | >>>>>> Also try running `kitchen diagnose --all` for configuration
```

### PR DESCRIPTION
**Pull request by tas50**
Fixes #606.

## What was actually wrong

The reporter's failure was a local OpenSSL/Digest fault, but what they were shown was this:

> Failed to complete #create action: [Digest initialization failed: initialization error **in the specified region**]

That "in the specified region" tail is ours. `#create` rescued `Exception` and re-raised a bare `RuntimeError` built as:

```ruby
raise "#{e.message} in the specified region #{config[:region]}. Please check this AMI is available in this region."
```

So **every** create failure — expired credentials, a missing subnet, a local OpenSSL problem — was reported as an AMI availability problem, and the original exception class and backtrace were thrown away. The reporter went looking at their AMI, which was never the issue.

Rescuing `Exception` also meant a Ctrl-C during create got caught, cleaned up, and re-raised as a `RuntimeError` — reporting the user's own interrupt as a driver failure.

## The change

- Raise `Kitchen::ActionFailed` instead of `RuntimeError`. Test Kitchen's `ActionRunner` rescues it specifically and adds the "see the instance log" pointer.
- Include the original exception class in the message, and preserve its backtrace.
- Append the AMI hint **only** when the failure is actually about the image — an EC2 error code beginning `InvalidAMI`, or a driver-raised message mentioning an AMI. The hint is worth keeping; an AMI missing from the region is a genuinely common mistake.
- Re-raise `SignalException` and `SystemExit` untouched, while still cleaning up first so an interrupted create can't leak a billable instance.

The broad rescue is kept deliberately — cleanup on interrupt is the point — and the constant is now root-qualified. There's no `Kitchen::Exception` today, but `Kitchen::StandardError` **does** exist, and an unqualified constant would silently pick one up if it were ever added. (That exact shadowing bug is real elsewhere in this gem; see #667.)

## Before and after

Same three failures, driven through the real `#create`:

| Failure | Before | After |
| --- | --- | --- |
| `ArgumentError: Digest initialization failed` | `Digest initialization failed: initialization error **in the specified region us-east-1. Please check this AMI is available in this region.**` | `Failed to create the EC2 instance: ArgumentError: Digest initialization failed: initialization error` |
| `InvalidAMIIDNotFound` | same AMI tail | `Failed to create the EC2 instance: Aws::EC2::Errors::InvalidAMIIDNotFound: The image id '[ami-…]' does not exist **Check that image ami-… exists and is available in region us-east-1.**` |
| `AuthFailure` | same AMI tail | `Failed to create the EC2 instance: Aws::EC2::Errors::AuthFailure: AWS was not able to validate the provided access credentials` |

## Scope

This fixes the driver's **reporting**, not the OpenSSL fault itself. That's a dependency problem — the issue thread traces it to `rubyntlm`, fixed in [rubyntlm 0.6.4](https://github.com/WinRb/rubyntlm/blob/master/CHANGELOG.md#064-2024-06-06). The reporter's later log shows converge succeeding, with a separate `logging`-gem finalizer `ThreadError` at process exit, which is also outside this gem. What was in our control was sending them to debug the wrong thing.

## Verification

This error path had **no test coverage**, which is how the behavior survived. Nine examples added; I confirmed all nine fail against the previous implementation:

```
=== new specs against the UNFIXED lib ===
9 examples, 9 failures
=== against the fixed lib ===
9 examples, 0 failures
```

`bundle exec rake test` — 243 examples, 0 failures.
`bundle exec cookstyle --chefstyle` — 24 files, 0 offenses.

Note for reviewers: #670 also touches this rescue (it extracts the body of `#create` so a pre-flight error isn't buried by this same rewriting). The two will conflict textually in `ec2.rb`; the intents compose cleanly, and whichever lands second is a small manual merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tas50:
Independent confirmation of this one, from running the driver against live EC2 rather than from the issue: launching into an account with **no default VPC** produced

```
Failed to complete #create action: [No default VPC for this user in the specified
region us-west-2. Please check this AMI is available in this region.]
```

The fix there is `subnet_id`, which the message never mentions, while the advice it does give sends you to the EC2 console to check an image that was fine. Same shape as #606 with a completely different underlying cause, so the change here holds up against a second real trigger. A `subnet_filter` value that YAML parsed as a boolean produced the same suffix bolted onto an SDK type error.

I'd opened #679... sorry, #680 for this before finding yours, and have closed it — this one is more complete.

**One gap worth folding in here:** `destroy(state)` is still unguarded, so a cleanup that raises replaces the error that caused it:

```ruby
destroy(state)   # if this raises, the original failure is gone
raise if e.is_a?(::SignalException) || e.is_a?(::SystemExit)
```

An auto-created security group that cannot be deleted (still in use, dependency violation) is a realistic way to hit this, and it leaves the user debugging a failed `delete_security_group` instead of the reason the instance never came up. Suggested:

```ruby
# A cleanup that fails is worth reporting, but it must not replace the
# error that caused the cleanup.
begin
  destroy(state)
rescue ::Exception => cleanup_error
  error("Failed to clean up after a failed create: #{cleanup_error.message}")
end
```

and a regression test, which fails without the above (the `RuntimeError` escapes instead):

```ruby
it "reports a cleanup failure without hiding the original error" do
  allow(driver).to receive(:destroy).and_raise(RuntimeError, "could not delete security group")
  allow(driver).to receive(:submit_server)
    .and_raise(::Aws::EC2::Errors::VPCIdNotSpecified.new(nil, "No default VPC"))

  expect { driver.create(state) }.to raise_error(Kitchen::ActionFailed, /No default VPC/)
  expect(logged_output.string).to include("could not delete security group")
end
```

Happy to push that to this branch if you'd like.

PR #680, opened 2026-08-23
fix: Stop reporting every create failure as an image problem

tas50:
## Problem

Every exception raised while creating an instance is replaced with a bare `RuntimeError` carrying the same advice, whatever actually went wrong:

```ruby
rescue Exception => e
  destroy(state)
  raise "#{e.message} in the specified region #{config[:region]}. Please check this AMI is available in this region."
end
```

Two things go wrong there:

1. **The advice is only true for one of the ways create fails.** It is appended unconditionally.
2. **`raise "string"` is not a re-raise.** It constructs a new `RuntimeError`, so the original exception class and its backtrace are discarded.

Launching into an account with no default VPC reports:

```
Failed to complete #create action: [No default VPC for this user in the
specified region us-west-2. Please check this AMI is available in this region.]
```

The image is fine. The fix is to set `subnet_id`, which the message gives no hint of — while the advice it *does* give sends the reader off to check AMIs in the EC2 console. A `subnet_filter` whose value YAML parsed as a boolean produced the same suffix bolted onto an SDK type error:

```
[expected params[:filters][0][:values][0] to be a String, got class TrueClass
instead. in the specified region us-west-2. Please check this AMI is available
in this region.]
```

Separately, `destroy(state)` on the way out could raise and **replace** the error that caused it, leaving the user debugging a failed `delete_security_group` rather than the reason the instance never came up.

## Fix

- Unrelated failures are re-raised untouched, keeping their class and backtrace.
- The advice is kept for the case it was written for — an EC2 error code of `InvalidAMIID*` — and now names the image it is talking about.
- `Exception#exception` is used to add that context. It copies the receiver, so the error class, the EC2 error code and the backtrace all survive. (Reconstructing via `e.class.new(msg)` would not work here: `Aws::Errors::ServiceError`'s constructor is `(context, message)`, so the message would land in the context slot.)
- A cleanup failure is logged and the original error is raised.

## Verification

Full unit suite passes, with new `#create` coverage (there was none before) for each behaviour above. Verified against live EC2 in `us-west-2`:

**No default VPC** — before:

```
[No default VPC for this user in the specified region us-west-2.
 Please check this AMI is available in this region.]
```

after:

```
[No default VPC for this user]
```

**Invalid image id** — after, still explains itself, and now names the image:

```
[Invalid id: "ami-00000000000000000". Check that image ami-00000000000000000
 is available in region us-west-2.]
```

## Note on `rescue Exception`

I left the `rescue Exception` as-is rather than narrowing it to `StandardError`. It is there so that interrupting a create still tears down the half-built instance instead of leaking a billable resource, which seems deliberate and worth keeping. It is now a genuine re-raise, so an `Interrupt` stays an `Interrupt` instead of being converted into a `RuntimeError` about AMI availability.


tas50:
Closing this — I opened it before spotting #671, which fixes the same bug and does it better: it wraps in `Kitchen::ActionFailed` so `ActionRunner` adds the instance-log pointer, passes `SignalException`/`SystemExit` through untouched, and includes the original exception class in the message. Mine did none of those.

For the record, this was found independently by running the driver against live EC2 — launching into an account with no default VPC reported:

```
[No default VPC for this user in the specified region us-west-2.
 Please check this AMI is available in this region.]
```

which is the same failure mode as #606, so #671's fix is confirmed against a second real-world trigger.

**One gap #671 does not cover**, which I'd suggest folding in there rather than keeping this open for: `destroy(state)` is still called unguarded on the way out, so a cleanup that raises *replaces* the error that caused it. A security group that cannot be deleted hides the reason the instance never came up. I've left a comment on #671 with the fix and a regression test.

PR #682, opened 2026-08-23
fix: Tunnel SSH over SSM using the AWS-StartSSHSession document

tas50:
## Problem

`use_ssm_session_manager: true` on its own hangs forever.

The proxy command the driver builds is:

```
aws ssm start-session --target i-... --region us-west-2
```

With no session document, `start-session` opens an **interactive shell session**, which speaks nothing SSH understands. Pointed at that as a `ProxyCommand`, SSH waits for a banner that never arrives, so `kitchen create` sits here until it is interrupted:

```
[AWS SSM Session Manager] SSM agent is available on instance i-076cb3b691f1dafdd
[AWS SSM Session Manager] Using proxy command: aws ssm start-session --target i-076cb3b691f1dafdd --region us-west-2
       Waiting for SSH service on ...
(hangs)
```

A document was only ever passed when `ssm_session_manager_document_name` was set:

```ruby
if driver_instance.config[:ssm_session_manager_document_name]
  cmd += ["--document-name", driver_instance.config[:ssm_session_manager_document_name]]
end
```

So the feature could not work in its default configuration — the user had to already know to name `AWS-StartSSHSession` themselves, which is not mentioned in the README or in `docs/ssm-session-manager.md`.

## Fix

Tunnelling SSH over SSM is what the `AWS-StartSSHSession` document is for, so it becomes the default. `portNumber=%p` is passed with it, because `Net::SSH::Proxy::Command` (which is what Test Kitchen's SSH transport builds from `state[:ssh_proxy_command]`) substitutes `%p` with the port being connected on — so a transport configured for a port other than 22 tunnels to that port rather than silently to 22.

An explicitly configured `ssm_session_manager_document_name` still wins.

## Verification

Full unit suite passes. The existing example asserted the old command exactly, which locked the bug in; it has been updated and joined by coverage for the port parameter and for not double-adding the document.

Confirmed by hand against a live Amazon Linux 2023 instance in `us-west-2` with an SSM instance profile, that the document is the difference:

```console
# without the document — hangs until killed
$ ssh -o ProxyCommand="aws ssm start-session --target i-... --region us-west-2" ec2-user@i-...

# with it
$ ssh -o ProxyCommand="aws ssm start-session --target i-... --region us-west-2 \
      --document-name AWS-StartSSHSession --parameters portNumber=%p" ec2-user@i-...
SSM-SSH-OK
ip-10-99-1-213.us-west-2.compute.internal
```

and end-to-end through `kitchen test` on the same setup: create, converge and destroy all complete, where before the run hung indefinitely after the agent came online.


tas50:
Related: #654 also touches this code path, but from the other direction — it documents \`AWS-StartSSHSession\` as a value a user *may* pass to \`ssm_session_manager_document_name\`. This PR makes it the default, so the feature works without the user having to know that name. The two are compatible; if #654 lands first this becomes a smaller change to the same \`else\` branch.

